PR #36, opened 2015-01-07
Remove WartRemover

puffnfresh:


dwijnand:
How come @puffnfresh?

folone:
Yeah, what is going on? Also in https://github.com/typelevel/typelevel.github.com/pull/35. Is there a discussion somewhere? /cc @milessabin @propensive @larsrh

non:
I just saw Brian's PR and figured that if he wanted WartRemover gone, I should remove it.

I assume this is fallout from the Scalaz/CoC situation, although I haven't spoken to Brian about it. I apologize if I'm assuming wrong.

non:
I will let @milessabin speak about #35, although I know that Paul Chiusano specifically requested that scalaz-stream be removed.

puffnfresh:
Just don't want to be associated with any of this.

puffnfresh:
But would still like to work on WartRemover, if I want to.

milessabin:
I won't say much here ... I don't want to preempt the announcement which will be coming shortly.

PR #37, opened 2015-01-07
Remove Argonaut from typelevel.github.com.

seanparsons:
This has been a hard and painful decision to make, but I'm afraid that Argonaut from this point onwards will not be associated with typelevel.scala.

### DIFF
--- a/index.md
+++ b/index.md
@@ -17,9 +17,9 @@ meta:
 
 <div class="row-fluid">
   <div class="span4">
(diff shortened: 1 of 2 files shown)


milessabin:
Understood. I don't consider bridges to have been burned and you'd be very welcome to return later once the dust has settled.

milessabin:
Actually, I'll redo do this, preserving alphabetical order.

PR #39, opened 2015-01-12
Remove scodec from website

mpilquist:
Scodec is moving to its own domain (scodec.org). Thanks for hosting the project for the past year!


non:
So long and thanks for all the codecs!

PR #40, opened 2015-01-24
Updated links to Shapeless and Spire GitHub repositories

propensive:
This is instead of the blank pages they currently point to.


larsrh:
Thanks, Jon!

propensive:
There's probably more I could do without too much effort, but this should fix the most glaring complaint!

PR #44, opened 2015-07-12
New post: "When are two methods alike?"

S11001001:
Second in a series on type parameters and members, for posting on July 16.

### DIFF
--- a/_posts/2015-07-16-method-equiv.md
+++ b/_posts/2015-07-16-method-equiv.md
@@ -44,9 +44,16 @@ import scala.collection.mutable.ArrayBuffer
 
 def copyToZero(xs: ArrayBuffer[_]): Unit =
   xs += xs(0)
+
+TmTp2.scala:9: type mismatch;
+ found   : (some other)_$1(in value xs)
+ required: _$1(in value xs)
+    xs += xs(0)
+            ^
 ```
 
-Likewise, the Java version has a similar problem.
+Likewise, the Java version has a similar problem, though the error
+message doesn’t give as good a hint as to what’s going on..
 
 ```java
 import java.util.List;
@@ -54,6 +61,10 @@ import java.util.List;
 void copyToZero(final List<?> xs) {
     xs.add(xs.get(0));
 }
+
+TmTp2.java:11:  error: no suitable method found for add(CAP#1)
+        xs.add(xs.get(0));
+          ^
 ```
 
 Luckily, in both Java and Scala, we have an *equivalent* method type,
@@ -74,13 +85,11 @@ private def copyToZeroP[T](xs: ArrayBuffer[T]): Unit =
 Similarly, in Java,
 
 ```java
-import java.util.List;
-
 void copyToZeroE(final List<?> xs) {
     copyToZeroP(xs);
 }
 
-void <T> copyToZeroP(final List<T> xs) {
+<T> void copyToZeroP(final List<T> xs) {
     final T zv = xs.get(0);
     xs.add(zv);
 }
@@ -107,9 +116,14 @@ crushing the inferred type of `zv` to `Any`.
 ```scala
 def copyToZeroE(xs: ArrayBuffer[_]): Unit = {
   val zv = xs(0)
-  // TODO error
   xs += zv
 }
+
+TmTp2.scala:19: type mismatch;
+ found   : zv.type (with underlying type Any)
+ required: _$1
+    xs += zv
+          ^
 ```
 
 When we call the type parameterized variant to implement the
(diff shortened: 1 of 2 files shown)


_posts/2015-07-16-method-equiv.md
```diff
@@ -0,0 +1,354 @@
...
+type parameter—or a method type parameter lowered to an
+existential—but there are other pairs of method types I want to
+explore that are the same, or very close.  So let’s talk about how we
+determine this equivalence.
+
+A method *r* is more general than or as general as *q* if *q* may be
+implemented by only making a call to *r*, passing along the arguments.
+By more general, we mean *r* can be invoked in all the situations that
+*q* can be invoked in, and more besides.  Let us call the result of
+this test *r* <:*ₘ* *q*, where <:*ₘ* is pronounced “party duck”; if
```
non: Maybe make the "where <:*ₘ* is pronounced 'party duck'" part parenthetical?

> Let us call the result of this test *r* <:*ₘ* *q* (where <:*ₘ* is pronounced “party duck”).

_posts/2015-07-16-method-equiv.md
```diff
@@ -0,0 +1,354 @@
...
+the test of *q* making a call to  *r* fails, then ¬(*r* <:*ₘ* *q*).
+
+If *q* <:*ₘ* *r* and *r* <:*ₘ* *q*, then the two method types are
+*equivalent*; that is, neither has more expressive power than the
+other, since each can be implemented merely by invoking the other and
+doing nothing else.  We write this as *q* ≡*ₘ* *r*.  Likewise, if
+*r* <:*ₘ* *q* and ¬(*q* <:*ₘ* *r*), that is, *q* can be written by
+calling *r*, but not vice versa, then *r* is *strictly more general*
+than *q*, or *r* <:*ₘ* *q*.
```
non: The way I read this, you are saying that there are two relations, *party duck* and *strictly-party duck*. But I see the same symbol (<:*ₘ*) being used for both. Should one be <≡:*ₘ*?

tpolecat:
Whatever character you have after the `<:` is just showing up as a rectangle for me. Is it supposed to be something else?

non:
@tpolecat I see a triple equals (congruence) symbol.

S11001001:
@tpolecat It's an italicized, subscripted "m", to draw an analogy with the "weak conformance" symbol from SLS, which uses a little "w" instead.  Can you tell me more about the font it's using?

_posts/2015-07-16-method-equiv.md
S11001001: @non Thanks, adopted that form.

_posts/2015-07-16-method-equiv.md
S11001001: @non No, but the example of the latter was wrong.  Fixed, thanks.

tpolecat:
Yeah it shows up as `LATIN SUBSCRIPT SMALL LETTER M` in the special chars dialog if i paste it in, but it doesn't appear to be present in any font that I have. It's using Helvetica Neue.

S11001001:
@tpolecat Including the one on typelevel.org posts?  It only really matters there, I suppose.  I'm loath to give up the symbol, but alternate encodings would be nice.

tpolecat:
Yeah it's the same font there. Maybe we could turn on the TeX formatting doodad that they use for SLS. I don't know how that works.

S11001001:
@tpolecat That might be pretty, but I'll try getting away with &lt;sub> and &lt;small> first, e.g. *<sub><small>m</small></sub>*.

(Sure, they work on GH, but jekyll is what matters.)

larsrh:
You can enable LaTeX markup by setting `mathjax: true` in the front matter of the page (just like `pygments: true`). There should be some examples in the existing posts. Also, it's in the README :-)

S11001001:
@larsrh Hah, I didn't notice that, thank you!

I'll experiment and share a screenshot here; see if you like how it is in mathjax better than it is now (@tpolecat it should look good to you now).

S11001001:
Q and R capitalized but left italic when standing alone, expressions translated to mathjax, DejaVu fonts.

http://i.imgur.com/sHnprNq.jpg

Compared to those paragraphs in the current draft, how does it look to you?

tpolecat:
Yep, looks ok now. I think the mathjax version is prettier and easier to read, although I would probably go ahead and encode the standalone Q/R.

S11001001:
OK, total mathjax has landed.

S11001001:
As usual, though, I've just α-converted until it works, and maybe these variable names in the definitions are less clear.

S11001001:
@tpolecat Do the new subscripts in 6d5ab21 look OK to you?

tpolecat:
@S11001001 looks great.

non:
:+1:

S11001001:
Published.  Thank you @non, @tpolecat, @larsrh for reviewing!

PR #46, opened 2015-07-19
New post: "Type projection isn't that specific"

S11001001:
Fourth in a series on type parameters and members, for posting on July 23.  Preceded by #45.


non:
I don't have any specific comments yet, but the article has left me with a strong sense of anticipation for the next one. Well done!

tpolecat:
Yep, this is good. Read through a few times, played with the code. Very clear. :+1:

PR #47, opened 2015-07-23
New post: "Nested existentials"

S11001001:
Fifth in a series on type parameters and members, for posting on July 27. Preceded by #46.


puffnfresh:
The backticks in paragraph "The reason you can’t invoke..." aren't rendering on GitHub - might have a Markdown problem there.

S11001001:
@puffnfresh Yeah, that happens in `div side-note` as corrected in 3c94791. A non-hack solution would be nice, though.

S11001001:
Here are the "full relationships", rendered with Mathjax:

tptm 5 derivations

Does this mishmash of `code` and $$ look good here?

S11001001:
Well, I did the hacky version in 3d18e85, @puffnfresh.

PR #49, opened 2015-07-28
New post: "Values never change types"

S11001001:
Sixth in a series on type parameters and members, for posting on July 30. Preceded by #47.


S11001001:
Posted; thank you @non for reviewing and the ref suggestion.

PR #50, opened 2015-08-05
New post: Machinist vs. value classes

larsrh:
Original author: @non 
Source: https://gist.github.com/non/a6ff3c0796e566db20d1

### DIFF
--- a/_posts/2015-08-06-machinist.md
+++ b/_posts/2015-08-06-machinist.md
@@ -1,0 +1,434 @@
+---
+layout: post
+title: Machinist vs. value classes
+
+meta:
+  nav: blog
+  author: non
+  pygments: true
+---
+
+Machinist vs. value classes
+===========================
+
+This article is about [machinist](https://github.com/typelevel/machinist), a stand-alone project which started out as part of the [spire](https://github.com/non/spire) project and has been originally published in [October 2014](https://gist.github.com/non/a6ff3c0796e566db20d1).
+The original description can be found on [this blog]({% post_url 2013-10-13-spires-ops-macros %}).
+You should read that linked post first if you are not familiar with how Machinist works.
+
+
+Introduction
+------------
+
+
+[Machinist Issue #2](https://github.com/typelevel/machinist/issues/2) asks:
+
+> Is it correct, that this stuff is completely obsolete now due to
+> value classes or are there still some use cases? An example of using
+> value class for zero-cost implicit enrichment: [...]
+
+The short answer is that Machinist value classes existed before the Machinist macros were implemented, and they do not solve the same problem Machinist solves.
+
+This article is the long answer.
+
+
+The base case
+-------------
+
+The Machinist's goal is to remove *any* overhead that would distinguish
+using a type class "directly" from using it indirectly via an implicit
+operator.
+
+Imagine we have the following "toy" type class:
+
+```scala
+trait Div[A] {
+  def div(lhs: A, rhs: A): A
+}
+
+object Div {
+  implicit val DivString = new Div[String] {
+    def div(lhs: String, rhs: String): String = lhs + "/" + rhs
+  }
+}
+```
+
+This allows us to write generic code such as:
+
+```scala
+class Test1 {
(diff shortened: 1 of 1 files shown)


_posts/2015-08-06-machinist.md
```diff
@@ -0,0 +1,434 @@
...
+------------
+
+
+[Machinist Issue #2](https://github.com/typelevel/machinist/issues/2) asks:
+
+> Is it correct, that this stuff is completely obsolete now due to
+> value classes or are there still some use cases? An example of using
+> value class for zero-cost implicit enrichment: [...]
+
+The short answer is that Machinist value classes existed before the Machinist macros were implemented, and they do not solve the same problem Machinist solves.
```
non: s/Machinist value classes/value classes/

larsrh:
@non If you don't object I'm going to merge soon-ish.

non:
yeah go for it :+1:

non:
i didn't notice other problems, if we have any we can fix them later.

larsrh:
Next one coming up in a couple of hours.

_posts/2015-08-06-machinist.md
```diff
@@ -0,0 +1,434 @@
...
+  def gen[A: Div](x: A, y: A): A = x / y
+  def test: String = gen("foo", "bar")
+}
+```
+
+Now, we can just say `x / y` and have that call `Div#div` automatically. We also don't need a reference to `ev: Div[A]` so we can use the nicer `[A: Div]` syntax. 
+
+With a normal implicit conversion, every call to `gen` would construct an instance of `Test3.DivOps`. However, since we have defined `Test3.DivOps` as a value class (by extending `AnyVal`), the object instantiation is ellided. Instead, the method call is dispatched to `Test3.DivOps.$div$extension` which calls `ev.div`.
+
+We often talk about value classes as not having a *cost*. Since no class is instantiation, we are not required to pay a cost in allocations, but we do still pay a cost in indirection (instead of calling `ev.div` directly as in `Test1` we have an intermediate extension method).
```
S11001001: s/instantiation/instantiated/

_posts/2015-08-06-machinist.md
```diff
@@ -0,0 +1,434 @@
...
+-------
+
+There are a few caveats that are worth mentioning:
+
+### Managing compilation units
+
+The issue that sparked this article used the operator `+/+`. Machinist claims to be able to support any symbolic operator. Why didn't we use that operator here?
+
+The answer has to do with how Scala macros work. Scala requires that macros be defined in a separate "compilation unit" from the one they are invoked in. This makes it very awkward to create a code snippet that both defines and uses a macro. In this case, it means that we can't extend `machinist.Ops` to define new symbolic operators in the same file that demonstrates their use. This is why we used `/` (which maps to `div` and is a "default operator").
+
```
S11001001: I would add a followup paragraph here, just to be entirely clear:

> You can arrange your "real" projects so that they are not affected by this limitation.

_posts/2015-08-06-machinist.md
```diff
@@ -0,0 +1,434 @@
...
+### Use outside of generic methods
+
+Now that we've demonstrated the cost that implicit conversions to value classes impose, you might imagine wanting to perform this transformation on *all* your implicit conversions.
+
+Unfortunately, Machinist is not sufficiently general to support this. Right now its macros support a number of different "shapes" but assume generic methods with type parameters. It might be possible to write macros which inline the method body of a concrete implicit class, but that's outside the scope of the project.
```
S11001001: This restriction, as written, seems to exclude `div`: it is not a "generic method with type parameters".  Can you say exactly what the restrictions are here?

_posts/2015-08-06-machinist.md
non: Yes, that's a great addition.

_posts/2015-08-06-machinist.md
non: Oh, yes, that should probably read *generic method which dispatches to an implicit evidence parameter*.

larsrh:
@S11001001 e9cf1bfba20081aae25f685d8582bf9796b26fba

Issue #32, opened 2014-10-13
Clarify "community venues" in the Code of Conduct


kpfell:
The current Code of Conduct states,
> We expect participants to follow these rules at all community venues (including online venues) and social events associated with the community.

We should clarify which venues and events this line refers to--the Github repos? Google Group? IRC channel? Other places not listed on the "About" page? What venues, specifically, fall under the Typelevel umbrella? We should provide a comprehensive list of (online _and_ offline) venues within the Code of Conduct itself.

Or, alternatively, if we created this list as a separate page within the website--then linked to it from the Code of Conduct--it would be helpful not only in enforcing the Code of Conduct, but in serving as a place to point newcomers who are seeking assistance.

adelbertc:
Looks like this is done via #53

PR #53, opened 2015-11-13
Clarify "community venues" in CoC

ceedubs:
This is a first pass at a resolution for #32. There is almost certainly
room for improvement and expansion. I wanted to get _something_ out
there, as this has been delayed for far too long.

@kpfell if you are still interested in this matter, I would greatly appreciate a review from you!

### DIFF
--- a/conduct.md
+++ b/conduct.md
@@ -22,11 +22,12 @@ If you are being harassed, notice that someone else is being harassed, or have a
 
 Community staff will be happy to help participants contact local law enforcement or otherwise assist those experiencing harassment to feel safe. We value your participation.
 
-We expect participants to follow these rules at all community venues (including online venues) and social events associated with the community. These venues include (but are not necessarily limited to): 
-* GitHub repositories and comments (specifically, all repositories under the [typelevel organization](https://github.com/typelevel))
-* Chat rooms (specifically, all Gitter channels under the [typelevel organization](https://gitter.im/orgs/typelevel/rooms) and the #typelevel IRC channel on Freenode)
-* Mailing lists (specifically, the [Typelevel Google Group](https://groups.google.com/forum/#!forum/typelevel))
-* Any conference or meetup that adopts the typelevel code of conduct (which includes all that are organized by the typelevel organization)
+We expect participants to follow these rules at all Typelevel-supported venues --including online venues-- and social events associated with Typelevel. These venues include but are not necessarily limited to: 
+* All GitHub repositories and comments under the [Typelevel organization](https://github.com/typelevel)
+* All Gitter channels under the [Typelevel organization](https://gitter.im/orgs/typelevel/rooms),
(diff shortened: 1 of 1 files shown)


milessabin:
:+1:

non:
:+1:

tpolecat:
:+1:

kpfell:
Hiya, @ceedubs!

Two suggestions:
1. Since this CoC is for Typelevel, I think it would be clearer to list the Typelevel locations first, then the more general category those locations fall into second--if at all.
2. In the first line, I'd rephrase "community venues" and "social events associated with the community," as I'm not sure they're specific enough.

For example,

```
We expect participants to follow these rules at all Typelevel-supported venues--including online venues--and social events associated with Typelevel. These venues include but are not necessarily limited to: 
* All GitHub repositories and comments under the [Typelevel organization](https://github.com/typelevel)
* All Gitter channels under the [Typelevel organization](https://gitter.im/orgs/typelevel/rooms),
* The #typelevel IRC channel on Freenode,
* The [Typelevel Google Group](https://groups.google.com/forum/#!forum/typelevel), and
* All conferences organized by the Typelevel organization.
```

It would also be useful if you could specify who/what is "the Typelevel organization," if possible.

tl;dr I think generalizing codes of conduct is useful in some cases, but given that you can't enforce the CoC outside of the repositories, chat rooms, etc. that you, yourselves, run, I'd leave the generalizations out for the time being.

ceedubs:
@kpfell thank you very much for the prompt, detailed, and helpful response.

It seems that we hitting an awkward point by trying to have a code of conduct that can be generally adopted by interested parties but that is specific enough to provide clear guidelines for typelevel projects/venues.

I think that your recommendations of being as specific and clear as possible are good ones (so long as it is noted that these are not necessarily all-encompassing). I will update this PR when I have a little more time, which I expect to be no later than Monday.

ceedubs:
I have pushed a new commit that basically directly copies the suggestion by @kpfell.

> It would also be useful if you could specify who/what is "the Typelevel organization," if possible.

I agree. I think right now Typelevel is really only defined in a few peoples' minds, and I don't think that all of those people necessarily have the exact same mental definition. Furthermore, I think the Typelevel website doesn't portray the current focuses of Typelevel very well. My thought here is that hopefully "who is Typelevel" will be better answered after #54, and that "what is Typelevel" is a question that we need to work on a bit and that probably shouldn't block having fairly well-defined "venues" in the CoC. How does that sound?

kpfell:
@ceedubs Ah, alright. :+1:, then!

milessabin:
:+1: to the updates.

ceedubs:
It looks like there are quite a few :+1:s on this and #54. FYI I don't have merge rights, so if these are ready to merge, someone else will need to do that.

milessabin:
Merging :-)

Issue #33, opened 2014-10-13
Clarify "community staff" in the Code of Conduct


kpfell:
It would be useful to list who, specifically, counts as a "member of the community staff." 

There's a link to the info@typelevel.org address, but I can see that being problematic if, for example, someone wanted to report a person who's well-known in the community. Without knowing who has access to the info@typelevel.org address, the reporter couldn't be certain that the person they're reporting wouldn't see the message first and retaliate.

That's problematic if the goal is to empower people to stop inappropriate behavior.

adelbertc:
Looks like this is done via #54

PR #54, opened 2015-11-13
Clarify "community staff" in CoC

ceedubs:
This is a first pass at a resolution to #33.

I've added email links for a few people that already had their email
addresses publicly available and that I felt confident would be happy to
be on this contact list. We can definitely add more contacts if there
are others who would be willing to be contacts.

### DIFF
--- a/conduct.md
+++ b/conduct.md
@@ -24,7 +24,7 @@ Community staff will be happy to help participants contact local law enforcement
 
 We expect participants to follow these rules at all community venues (including online venues) and social events associated with the community.
 
-For any questions or with any concerns at all, please contact any member of community staff. Specifically, you can contact:
+If you have any questions or concerns, please contact a member of the community staff, such as:
 * [Miles Sabin](mailto:%6d%69%6c%65%73@%6d%69%6c%65%73%73%61%62%69%6e.%63%6f%6d)
 * [Erik Osheim](mailto:d_m@plastic-idolatry.com)
 * [Stew O'Connor](mailto:%73%74%65%77@%76%69%72%65%6f.%6f%72%67)
--- a/conduct.md
+++ b/conduct.md
@@ -24,7 +24,12 @@ Community staff will be happy to help participants contact local law enforcement
 
 We expect participants to follow these rules at all community venues (including online venues) and social events associated with the community.
 
-For any questions or with any concerns at all, please contact any member of community staff.
+For any questions or with any concerns at all, please contact any member of community staff. Specifically, you can contact:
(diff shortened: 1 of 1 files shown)


milessabin:
:+1:

non:
:+1:

tpolecat:
:+1:

adelbertc:
:+1:

ceedubs:
@kpfell as you were so helpful with #53, do you have any input on this as well?

larsrh:
Just want to add that there is also the general @typelevel/owners address posted on the [about page](http://typelevel.org/about.html), although it might be better to list personalized addresses.

stew:
:+1:

kpfell:
@ceedubs I'd rephrase this slightly to address the reader, as a form of encouragement:

> If you have any questions or concerns, please contact a member of the community staff, such as [...]

Since you have the email address that @larsrh mentioned, as well, I would definitely tack a note onto the end about it--though, if a group of people has access to that account, I would mention who those people are.

ceedubs:
A concern that I have here is that documentation is prone to going out of date. We could list the current people that get `info@typelevel.org` emails, but that could change in the future, and if we forget to update the Code of Conduct accordingly, I would hate for an email to go to someone that it wasn't intended to go to. I could leave a note that the members of this email address could change over time, but that might not be that helpful to someone who wants to be sure about who will see their email.

What do you think?

ceedubs:
By the way, @kpfell, I do like like the suggestion of addressing the reader for encouragement and will definitely incorporate that change regardless of what's decided about the Typelevel email address.

non:
Let's just list people explicitly. In addition to avoiding the out-of-date documentation problem, it also makes it easy for someone to avoid including someone they don't want to.

For example, if someone wanted to report me for a CoC violation it would be bad if they sent it to info@ without realizing that I would be included.

non:
:+1:

kpfell:
@ceedubs 

> We could list the current people that get info@typelevel.org emails, but that could change in the future, and if we forget to update the Code of Conduct accordingly, I would hate for an email to go to someone that it wasn't intended to go to. I could leave a note that the members of this email address could change over time, but that might not be that helpful to someone who wants to be sure about who will see their email.

Ah, okay--fair enough! I wasn't sure if you had a GitHub team set up that corresponded with people able to access the account.

:+1: if not. :wink:

milessabin:
:+1:

ceedubs:
@kpfell I'm not sure exactly how the `info@typelevel.org` email address is set up. Paging @larsrh.

larsrh:
@ceedubs It's an alias on my personal mail server.

ceedubs:
@larsrh okay thanks. I think we can stick with what we have for now then.

ceedubs:
I think this is good to merge.

PR #58, opened 2015-12-22
Add CfP to homepage

d6y:
I found it surprisingly hard to find the CfP, so thought maybe others would too.  This PR adds the CfP on the home page:

typelevel

Just an idea. It's ok --  I'm not precious about it. Feel free to close or re-work if it's not really what's wanted :-)


milessabin:
:+1: ... @larsrh do you want to do this now or wait for the site refresh?

larsrh:
Can't hurt to have it on the homepage already :smile: Thanks, @d6y!

Issue #68, opened 2016-01-20
Event images should be centered on home page


propensive:
These are currently shifted about 20px to the right, which offsets them from the centered text beneath, and (noticeably) from the edges of the viewport when viewing on a phone.

PR #69, opened 2016-01-20
Centered event images on home page

propensive:
Fixes issue #68


larsrh:
Since I don't speak CSS I'm assuming you did the right thing. Thanks!

PR #74, opened 2016-01-22
New post: "It's existential on the inside"

S11001001:
Eighth in a series on type parameters and members, for posting on January 28. Preceded by #52.

I hope you like the pictures.

NB: Links to other posts will not work on GH, because they use `post_url`, which only works on the site.


tpolecat:
:+1: very clear. Extra points for using the word *cromulent*.

larsrh:
With bonus drawings! :sparkles:

PR #77, opened 2016-01-26
New blog post: Of variance and functors

adelbertc:


_posts/2016-01-26-variance-and-functors.md
```diff
@@ -0,0 +1,281 @@
...
+A type that cannot safely be made covariant is `Array`. If `Array` were covariant, we could substitute
+an `Array[Circle]` for an `Array[Shape]`. This can get us in a nasty situation.
+
+```scala
+val circles: Array[Circle] = Array.fill(10)(Circle(..))
+val shapes: Array[Shape] = circles // works only if Array is covariant
+shapes(0) = Square(..) // Square is a subtype of Shape
+```
+
+If `Array` was covariant this would compile fine, but fail at runtime. The compiler would accept this
```
larsrh: Could you note here that in Java this compiles, but throws `ArrayStoreException` at runtime?

adelbertc: Done!

_posts/2016-01-26-variance-and-functors.md
```diff
@@ -0,0 +1,282 @@
...
+to show a `Circle` (`Show[Circle]`), we can give instead a way to show just a `Shape`. This is a valid
+substitution because we can show a `Circle` by throwing away `Circle`-specific information and showing just
+the `Shape` bits. This means that `Show[Shape]` is a subtype of `Show[Circle]`, despite `Circle` being a
+subtype of `Shape`.
+
+In general, we can show (or write) a subtype if we know how to show a supertype by tossing away subtype-specific
+information (an upcast) and showing the remainder. Again, this means `Show[Supertype]` is substitutable, or a
+subtype of, `Show[Subtype]`.
+
+For similar reasons that read-only types can be made covariant, write-only types can be made contravariant.
```
ceedubs: This may be a bit confusing. By "types" are you referring to the type class? We've shown that we can both "read" and "write" a `Shape` so "read-only" and "write-only" don't seem to apply to `Shape` itself.

_posts/2016-01-26-variance-and-functors.md
```diff
@@ -0,0 +1,282 @@
...
+  new Invariant[Array] {
+    def imap[A, B](fa: Array[A])(f: A => B)(g: B => A): Array[B] =
+      new Array[B] {
+        // Convert read A to B before returning - covariance
+        override def read(i: Int): B =
+          f(fa.read(i))
+
+        // Convert B to A before writing - contravariance
+        override def write(i: Int, b: B): Unit =
+          fa.write(i, g(a))
```
ceedubs: Here we are mutating the initial array. This is dubious at best. Should we be using an immutable type instead? In the past when I've explained `Invariant` to people I've used `Invariant[Semigroup]` in an example, converting a `Semigroup[Long]` to a `Semigroup[Date]` (or something like that).

ceedubs:
@adelbertc thanks for writing this up! It will be nice to have a great blog post to link people to when this comes up in the future.

_posts/2016-01-26-variance-and-functors.md
adelbertc: I was referring to `List`, `Read`, and `Show` - maybe I should say type constructor instead? e.g. "For similar reasons that read-only type constructors ..." ?

_posts/2016-01-26-variance-and-functors.md
adelbertc: Here I was trying to explain why `Array` must be invariant since I was talking about how it couldn't be co- or contra-variant early on. Good call on adding an example that doesn't require mutation though, maybe to keep it consistent I'll write a `Serializer` type class that combines both `Read` and `Show`:

```scala
trait Serializer[A] extends Read[A] with Show[A]
```

_posts/2016-01-26-variance-and-functors.md
```diff
@@ -0,0 +1,309 @@
...
+```scala
+def upcast[F[_], A, B <: A](functor: Functor[F], fb: F[B]): F[A] =
+  functor.map(fb)(b => b: A)
+```
+
+`upcast`'s behavior does exactly what covariance does - given some supertype `A` (`Shape`) and a subtype `B` (`Circle`),
+we can mechanically (and safely) turn an `F[B]` into an `F[A]`. Put differently, anywhere we expect an `F[A]` we can provide
+an `F[B]`, i.e. covariance. For this reason, `Functor` is sometimes referred to in full as covariant functor.
+
+# Contravariance
```
S11001001: You might suggest that contravariance (of either sort) most often applies for datatypes that represent computations; this is because the only primitive-contravariant position is the fucntion argument. For example, the input to a Kleisli arrow, and [the "haystack" (`S`) to a fold](https://github.com/julien-truffaut/Monocle/blob/8efd6c85f8a3697abc11feae0a5f1a2ba7fd0a58/core/src/main/scala/monocle/Fold.scala#L28), are contravariant.

_posts/2016-01-26-variance-and-functors.md
```diff
@@ -0,0 +1,309 @@
...
+```scala
+def contraUpcast[F[_], A, B >: A](contra: Contravariant[F], fb: F[B]): F[A] =
+  contra.contramap(fb)((a: A) => a: B)
+```
+
+Going back to `Shape`s and `Circle`s, we can show a `Circle` by upcasting it into a `Shape` and showing that.
+
+# Invariance
+Unannotated type parameters are considered invariant - different instantiations of a type constructor have no
+relationship with one another regardless of what relationships the type parameters may have. Given invariant
```
S11001001: There is a type-relationship equivalent to invariant functors, though it depends on what you mean by 'different'. It's that type equality is congruent.

A = B → F[A] = F[B]

That's not nothing; GADT pattern matching reveals type equalities that only exist in certain blocks.

S11001001:
The fourth variance:

Unfortunately, Scala's variance model is incomplete in two ways. Here's # 1: 'phantom' is the fourth variance, aka "anyvariance" (thanks @ekmett). To explain, here are the four variances characterized as implications (assume "for all A, B"):

1. invariance: A = B → F[A] = F[B]
2. covariance: A <: B → F[A] <: F[B]
3. contravariance: A >: B → F[A] <: F[B]
4. phantom: F[A] = F[B]

This completes the [system of kind subsumption](https://github.com/scala/scala/pull/4666/files) alluded to in work surrounding the SI-2066 fix.

```
   invariance
     ↑   ↑
    /      \
   -        +
   ↑        ↑
    \      /
    phantom
```

You might show here that `trait Phantom[F[_]] extends Functor[F] with Contravariant[F]` (it has a method `[A, B](F[A])F[B]`).  To go even further, you might show that having both `Functor` and `Contravariant` always implies `Phantom`.

mpilquist:
Not sure if this is useful or not, but I explored phantom (aka bivariance) in these slides: http://mpilquist.github.io/blog/2014/05/27/variance/

I found the correspondence between functor composition and subtyping variance interesting.

_posts/2016-01-26-variance-and-functors.md
```diff
@@ -0,0 +1,309 @@
...
+before handing it over.
+
+Similarly if we make it write-only (like a write-only handle on a resource) we can safely treat
+it as contravariant. If we are asked to store `Circle`s and we know how to store `Shape`s,
+we can upcast each `Circle` into a `Shape` before storing it.
+
+Variance manifests in two levels: one at the type level where subtyping relationships are defined, and
+the other at the value level where it is encoded as an interface which certain types can conform to.
+The types that can implement `Functor` are precisely those that can be made covariant, and the types that
+can implement `Contravariant` are precisely those that can be made contravariant.
```
S11001001: I would not say this; it would be easiest to show a counterexample in both directions. I'll focus on `Functor`, but the same problems I discuss apply to `Contravariant` and `Phantom` too.

## Covariant parameter, not functor: openly subclassable

```scala
trait Has[+A] {
  def it: A
}

// in some downstream module
trait Saves[A] extends Has[A] {
  var it: A
}
```

Whatever your `map` does, it can't return a `Saves`, unless you change `Has` to include a subtyping-polymorphic `map`, at which point it isn't the same type anymore. So `Functor` on `Has` can't meet the identity law. (`Seq` et al have the same problem.)

Side-effects introduce a lot more problems in this category, but they aren't necessary for it to be a problem; open subclassing is sufficient to break the relationship.

## Functor, not covariant parameter: higher-kind without variance variable

In this type,

```scala
final case class FA[F[_], A](fa: F[A])
```

the variance of `A` depends on the variance of `F`. Variance subsumption (à la SI-2066) means that we can pass covariant, contravariant, or even phantom (supposing it existed) type constructors as `F`. However, for that to carry over to the variance of `A`, you would need "variance variables" (think a specialization of kind variables; the variance of a type parameter position is an element of the kind of the type constructor surrounding it).

This is the # 2 thing missing from Scala's variance models: variance quantification.

Meanwhile, it is trivial to model this relationship with typeclasses: `Functor[F]` → `Functor[FA[F, ?]]`, `Contravariant[F]` → `Contravariant[FA[F, ?]]`, and `Phantom[F]` → `Phantom[FA[F, ?]]`

_posts/2016-01-26-variance-and-functors.md
larsrh: > In the past when I've explained Invariant to people I've used `Invariant[Semigroup]` in an example, converting a `Semigroup[Long]` to a `Semigroup[Date]` (or something like that).

Isn't that a bit confusing, given that `Semigroup` itself is a type class? (Not saying it's a bad example, but probably better for explaining that type classes are just types in Scala, not for how invariance works.)

adelbertc:
This is what I get for trying to blog about fancy things :P 

Thanks for all the input, I'll read through the new resources and make some changes/additions :-)

adelbertc:
I haven't forgotten about this, just had a busy weekend. Will get to it soon!

larsrh:
While you're at it, could you please rename the file to have a more recent date?

adelbertc:
Ready for review once more!

_posts/2016-02-03-variance-and-functors.md
```diff
@@ -0,0 +1,475 @@
...
+Variance allows us to define the subtyping relationships between type constructors - that is, under which
+conditions `F[A]` is a subtype of `F[B]`.
+
+Similarly in functional programming, there are covariant functors, contravariant functors, and invariant functors. The
+similarity in names is not coincidental.
+
+# Covariance
+The common example is `List[+A]` which is covariant in its type parameter, denoted by the `+` next to the `A`.
+A type constructor with a covariant type parameter means that if there is a subtyping relationship between the
+type parameter, there is a subtyping relationship between the two instances of the type constructor. This means that
```
larsrh: s/This means that/For example/

("means that" already appears in the previous sentence)

_posts/2016-02-03-variance-and-functors.md
```diff
@@ -0,0 +1,475 @@
...
+          f(fa.read(i))
+
+        // Convert B to A before writing - contravariance
+        override def write(i: Int, b: B): Unit =
+          fa.write(i, g(a))
+      }
+  }
+```
+
+## SerDe
```
larsrh: `SerDe`?

larsrh:
Pet peeve: Could you replace `-` by `–`? Apart from that, :+1:, and thanks for the changes!

larsrh:
Thanks, @adelbertc!

PR #82, opened 2016-01-29
Sponsorship for Philadelphia

larsrh:
Review by @milessabin


milessabin:
Rather than have the no-logo sponsors show up as fairly small text next to the large logos it might be better to have a short paragraph above platinum saying something along the lines of "We'd like to thank all our sponsors, some of which appear below. We would also like to thank Verizon (Gold sponsor) and Comcast (Silver) ... updates to follow."

larsrh:
Updated with one more logo and design help from Vic.

PR #84, opened 2016-02-01
use Bundler to bootstrap Jekyll & related changes

larsrh:
Courtesy of ensime/ensime.github.io#24 and http://markdotto.com/2014/09/25/sass-and-jekyll/.


larsrh:
@fabe I had to move the `main.scss` file from `_scss` to `css` because otherwise, Jekyll's SCSS compiler wouldn't pick it up. The includes are still where they used to be though. The Gulp setup is probably broken now, but otherwise everything else should work with a simple `bundler` invocation.

dwijnand:
Hehe, the love is shared. Note things have upgraded in GitHub Pages today: https://github.com/blog/2100-github-pages-now-faster-and-simpler-with-jekyll-3-0

larsrh:
Yeah, learned that the hard way today, e.g. c7c81efad2a46b425392bcfe96d8d52fd3d1122b. The update has been deployed in the middle of a series of commits, which made everything even more confusing :shipit:

PR #89, opened 2016-02-07
Fix mistake with type bound

joiskov:


larsrh:
I don't think this change is correct. Consider the [Wikipedia article](https://en.wikipedia.org/wiki/Covariance_and_contravariance_%28computer_science%29#Function_types):

    S1 → S2 ≤ T1 → T2 if T1 ≤ S1 and S2 ≤ T2.

To me it looks like the @adelbertc's reproduces that accurately.

S11001001:
@larsrh right, the change is not right.

joiskov:
Okay. He wrote this
` Input[Shape] <: Input[Circle]`
but "An upper bound restriction states that any type selected must be equal to or a lower than the upper bound type".
How can this be correct?

larsrh:
I'm afraid I don't follow. Where are you quoting this from? It doesn't appear in the blog post nor in the Scala specification.

> `Input[Shape] <: Input[Circle]`

This is not an upper bound restriction, it represents the subtyping relationship between two types. Essentially: `Input[Shape]` is a subtype of `Input[Circle]`.

joiskov:
The quote from the book "Scala in Depth", Joshua Suereth, 2012 (page 132, last paragraph).
You are right! 

> Essentially: Input[Shape] is a subtype of Input[Circle].

but in example we have:

```scala
trait Shape
trait Circle with Shape
```

or am I mistaken?

larsrh:
Yes. That's precisely the point of contravariance. The subtyping relationship between `F[A]` and `F[B]` is _opposite_ to the relationship between `A` and `B`. It's really counter-intuitive when first seeing it.

I would suggest continuing the discussion in the comments of the [post](http://typelevel.org/blog/2016/02/04/variance-and-functors.html). Would be more than happy to help explaining in more detail there.

adelbertc:
@joiskov `Circle` is a subtype of `Shape` like you have in the example, but function parameters are contravariant, so a subtype of `Circle => Unit` is `Shape => Unit`. The type alias tries to make it more apparent:

```scala
scala> :pa
// Entering paste mode (ctrl-D to finish)

trait Shape
trait Circle extends Shape

type Input[A] = A => Unit
def shapeInput: Input[Shape] = ??? // implementation not important, we just want type checking1
def circleInput: Input[Circle] = shapeInput

// Exiting paste mode, now interpreting.

defined trait Shape
defined trait Circle
defined type alias Input
shapeInput: Input[Shape]
circleInput: Input[Circle]
```

We see that we can use a value of type `Input[Shape]` in a position where `Input[Circle]` is expected - that is, `Input[Shape] <: Input[Circle]`.

adelbertc:
@joiskov Please let me know at what point the blog post's discussion around contravariance got unclear, we'll need to fix that :-)

joiskov:
@adelbertc watch here http://typelevel.org/blog/2016/02/04/variance-and-functors.html#comment-2501607538

joiskov:
Thank you. I was wrong. @adelbertc right :)

PR #93, opened 2016-03-10
New blog post: Information hiding, enforced

adelbertc:


_posts/2016-03-10-information-hiding.md
```diff
@@ -0,0 +1,334 @@
...
+```scala
+trait Universal {
+  def apply[A]: A => A
+}
+```
+
+Here `Universal#apply` says *for all* choices of `A`, a function `A => A` can be
+written. In the [Curry-Howard Isomorphism][propositionsAsTypes], a profound
+relationship between logic and computation, this translates to "for all propositions
+`A`, `A` impies `A`." It is therefore acceptable to write the following, which picks
```
fthomas: s/impies/implies/

larsrh:
@S11001001, this one should be right up your alley. Mind having a look?

_posts/2016-03-10-information-hiding.md
```diff
@@ -0,0 +1,334 @@
...
+
+For example, a set with fast random indexing (useful for random
+walks on a graph) can be implemented with a sorted `Vector`.
+However, if the `Vector` type is
+leaked, the user can use this knowledge to violate the invariant.
+
+```scala
+import scala.annotation.tailrec
+
+def binarySearch(i: Int, repr: Vector[Int]): (Boolean, Int) = {
```
larsrh: I think the concrete binary search implementation here is not strictly relevant. Might be a bit easier to read if you elide the function body.

_posts/2016-03-10-information-hiding.md
```diff
@@ -0,0 +1,334 @@
...
+work with the type to the extent the module allows. This is precisely
+the notion of information hiding. If this can be enforced by the type
+system, modules can be swapped in and out without worrying about breaking
+client code.
+
+# Quantification
+It turns out there is a [well understood principle][understandingTypes]
+behind this idea called *existential quantification*. Contrast with
+universal quantification which says "for all", existential quantification
+says "there exists."
```
larsrh: I usually prefer to say "there is a" instead of "there exists", because otherwise you'd explain "existential" by "exists". (That's just my 2¢ as a non-native speaker.)

_posts/2016-03-10-information-hiding.md
```diff
@@ -0,0 +1,334 @@
...
+```
+
+Here `Universal#apply` says *for all* choices of `A`, a function `A => A` can be
+written. In the [Curry-Howard Isomorphism][propositionsAsTypes], a profound
+relationship between logic and computation, this translates to "for all propositions
+`A`, `A` implies `A`." It is therefore acceptable to write the following, which picks
+`A` to be `Int`.
+
+```scala
+def intUniversal(u: Universal): Int => Int =
```
larsrh: This function is not "universal" any more. How about `intInstantiation`?

_posts/2016-03-10-information-hiding.md
adelbertc: So one reason I left it on is because I'm running this all through `tut` so the code needs to be present to "pass." I do agree it's clutter though so I'll remove it after running it through tut

larsrh: You could replace it by `???` :zap:

adelbertc: It needs the actual implementation so Tut can generate the (commented)
output - I just commented out the implementation after Tut generation since
I do think it gets in the way of the more important stuff

larsrh:
:+1: from me

S11001001:
Content :+1:

Please check whether Jekyll supports your link syntax before merging.

adelbertc:
All links appear to work, modulo the ACM links, but those seem to be down for everyone right now (at least during the time I wrote this): http://www.downforeveryoneorjustme.com/http://dl.acm.org/citation.cfm?id=6042

adelbertc:
Looks like ACM links are working now!

larsrh:
Merged, thanks. @adelbertc, please subscribe to the comment thread.

adelbertc:
Thanks @larsrh and @S11001001 for reviewing! I've subscribed to the comment thread.

PR #96, opened 2016-03-25
remove the boulder summit, add a blog post explaining why

stew:


rossabaker:
I have seen more good thoughts that would make fine concurring opinions in a personal space.  In the organization's space, I find that this says what needs to be said and no more. :+1:

non:
I agree with @rossabaker

_config.yml
```diff
@@ -4,7 +4,7 @@ twitter: typelevel
 email: info@typelevel.org
 gitter: "https://gitter.im/orgs/typelevel/rooms"
 google_groups: "https://groups.google.com/group/typelevel"
-baseurl: http://typelevel.org
+baseurl: /
```
tpolecat: Was this change intentional?

tpolecat:
This whole thing is sad, but I agree it's what we should do.

_config.yml
stew: oh certainly no, thank you

milessabin:
Also :+:.

PR #99, opened 2016-05-27
Reworked events page

larsrh:
Comments welcome! The plan is to add the [Hack the Tower event](http://www.meetup.com/london-scala/events/231404561/) too, while still allowing convenient access to material from the past events.

/cc @milessabin 

## New front page

screenshot_20160527_222355

## New events page

screenshot_20160527_222403


milessabin:
:+1:

PR #100, opened 2016-05-28
Link to the /events page in the footer

fthomas:
... instead of the Philadelphia summit


larsrh:
Excellent catch!

larsrh:
Also, PR #100 :tada:

PR #108, opened 2016-07-24
Update head to include the RSS link

alexandru:
The RSS feed is not automatically discoverable because the <head> section does not indicate it. And it should, it makes discovery easier - for example Firefox's "Subscribe to this page" button does not work without it.


larsrh:
Good catch!

PR #111, opened 2016-08-03
New event: scala.io

larsrh:
cc @jeantil


jeantil:
Looks good to me !

Thank you :)

PR #112, opened 2016-08-16
Update org in catalysts github link

richardmiller:


milessabin:
Good spot. Many thanks ... merging :-)

PR #113, opened 2016-08-20
New post: "Higher-kinded types: the difference between giving up, and moving forward"

S11001001:
Thanks to a suggestion by @sethtisue about having more blogs out there motivating HKTs.


adelbertc:
Ohay https://scalaebythebay2016.sched.org/event/7iTp/spoiled-by-higher-kinded-types

tpolecat:
👍  Great!

SethTisue:
> Thanks to a suggestion by @sethtisue

heh, I don't remember this, but, good post!

a few nits:

* the names `Bindic` and `BindicTM` are mysterious
* "Note that we must use higher kinds to support our higher kinds" is cryptic — I *think* I know what you mean, but it could be explained a little more explicitly?

S11001001:
> the names Bindic and BindicTM are mysterious

The convention comes from [the ermine-parsers example, linked later](https://github.com/ermine-language/ermine-parser/blob/cc77bf6e150a16129744d18d69022f7b5902814f/src/main/scala/scalaparsers/Monadic.scala); the name `Monadic` is slightly inaccurate, but the name of the proper typeclass, `Bind`, doesn't admit a great trait variant name. `Bindable`? `SemiMonadic`?

> "Note that we must use higher kinds to support our higher kinds" is cryptic — I think I know what you mean, but it could be explained a little more explicitly?

Agreed; hope 1c704647be0dae4659dd5ea62ccf0434932c5760 suits a little better.

SethTisue:
`Bindable` seems to fine. 👍 to 1c70464

S11001001:
`Bindable` it is. 5df1a168f3ec87f7f1e88584c04dddd463b092c8

PR #121, opened 2016-10-09
replace Google Groups by Twitter on About page

larsrh:
## New rendering

screenshot_20161009_143221


larsrh:
Reason: There's very little activity on the mailing list anyway, and I realized that the home page doesn't even link to our Twitter account.

PR #136, opened 2016-12-09
The eff-cats library has been renamed to eff

etorreborre:
Thanks.


_data/incubator.yml
```diff
@@ -7,9 +7,9 @@
 - title: "delta"
   description: "Type-safe, generic diffing of data"
   github: "https://github.com/stacycurl/delta"
-- title: "eff-cats"
-  description: "Extensible effects for Cats"
-  github: "https://github.com/etorreborre/eff-cats"
+- title: "eff"
+  description: "Extensible effects for Scala"
+  github: "https://github.com/etorreborre/eff"
```
fthomas: Shouldn't that be https://github.com/atnos-org/eff? https://github.com/etorreborre/eff redirects to https://github.com/atnos-org/eff-scalaz which is marked as deprecated.

etorreborre: Thanks I didn't notice that :-).

etorreborre: I fixed that in the last commit.

PR #143, opened 2017-02-11
New post: 'There are more types than classes'

S11001001:
For posting on Monday, Feb 13.

### DIFF
--- a/posts/2017-02-13-more-types-than-classes.md
+++ b/posts/2017-02-13-more-types-than-classes.md
@@ -9,14 +9,14 @@ meta:
 ---
 
 As programmers, we are very incautious with our use of the word
-"type". The concept of "type" is sufficiently abstract and specific
+“type”. The concept of “type” is sufficiently abstract and specific
 that we are tempted to understand it by analogy, so much that we begin
 to confuse analogy with sameness.
 
-The colloquial "runtime type", a fair approximation of "class", makes
-it tempting to equate _types_ with "classes, interfaces, traits, that
-sort of thing", which I will name _classes_ for the rest of this
-article. But they aren't the same. The type system is much richer and
+The colloquial “runtime type”, a fair approximation of “class”, makes
+it tempting to equate _types_ with “classes, interfaces, traits, that
+sort of thing”, which I will name _classes_ for the rest of this
+article. But they aren’t the same. The type system is much richer and
 more interesting than the class system, even in Java.
 
 To appreciate this richness, we must stop thinking of types as classes
@@ -58,7 +58,7 @@ res3: (String, java.io.Serializable, CharSequence,
   (hi there!,hi there!,hi there!,hi there!,hi there!)
 ```
 
-So we have exhausted the classes, but aren't quite done with types.
+So we have exhausted the classes, but aren’t quite done with types.
 
 ```scala
 scala> greeting: greeting.type
@@ -97,16 +97,16 @@ scala> implicitly[greeting.type =:= salutation.type]
 <console>:14: error: Cannot prove that greeting.type =:= salutation.type.
 ```
 
-Now we have seven. I'll spare you spelling out the induction: each new
+Now we have seven. I’ll spare you spelling out the induction: each new
 variable defined like `salutation` will yield a new alias with a
 distinct type. This is not about _objects_; this is about _variables_!
 
 As local variables are a strictly compile-time abstraction, and we
-have anyway seen that the numbers don't match up, that should be the
-end of the "types are classes" confusion for you. But maybe this is
-just some Scala oddity! And anyhow I haven't even begun to demonstrate
+have anyway seen that the numbers don’t match up, that should be the
+end of the “types are classes” confusion for you. But maybe this is
+just some Scala oddity! And anyhow I haven’t even begun to demonstrate
 the overwhelming richness of the type model as it blindingly outshines
-the paucity of the class model. Let's go further.
+the paucity of the class model. Let’s go further.
 
 ## No values, infinite types: method type parameters
 
@@ -123,9 +123,9 @@ It seems like `G` must be `String`, because the argument passed to
 `pickGreeting` is a string, and in that case so must its return value
 be, according to the implementation. And from the perspective of this
 call, [outside]({% post_url 2016-01-28-existential-inside %})
-`pickGreeting`'s implementation, it is `String` indeed.
+`pickGreeting`’s implementation, it is `String` indeed.
 
-But that implementation's perspective matters, too; it is also part of
+But that implementation’s perspective matters, too; it is also part of
 our program. And it sees things quite differently. We can ask its
 thoughts on the matter by adding to its body
 
@@ -160,7 +160,7 @@ conformance[String, G]
                     ^
 ```
 
-Let's apply the pigeonhole principle. Imagine that you had a list of
+Let’s apply the pigeonhole principle. Imagine that you had a list of
 every class that ever was or ever will be. Imagine that, somehow, all
 of these classes, from `String` to
 `AbstractFactoryMethodProxyBuilder`, were on your classpath, available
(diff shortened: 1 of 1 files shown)


tpolecat:
This is fantastic. I'm going to put the URL on a hotkey. Didn't see any typos, although I'm kind of sleepy. 👍 👍 👍

posts/2017-02-13-more-types-than-classes.md
```diff
@@ -0,0 +1,702 @@
...
+
+```scala
+scala> conformance[salutation.type, String]
+
+scala> implicitly[greeting.type =:= salutation.type]
+<console>:14: error: Cannot prove that greeting.type =:= salutation.type.
+```
+
+Now we have seven. I’ll spare you spelling out the induction: each new
+variable defined like `salutation` will yield a new alias with a
```
non: You might want to clarify that you are using *variable* here in the sense of binding (`val`), not in the sense of mutation (`var`). Maybe this is overly pedantic since the examples all use `val` but I could imagine someone getting the wrong idea.

posts/2017-02-13-more-types-than-classes.md
```diff
@@ -0,0 +1,702 @@
...
+val greeting3 = later(3){
+  n => List.fill(n)("hi").mkString(" ")
+}
+```
+
+How many _classes_ are involved here, in the type of `greeting3`?
+
+1. `Later`, obviously;
+2. `String`, the output type of `greeting3.f`;
+3. `Int`, the `I` type.
```
non: Again this might be overly pedantic, but you might also want to mention the `Function1` class, which is involved in the type of `greeting3.f` as well.

non:
Great article! I had a few pedantic comments -- which you are free to ignore if you don't think they add anything useful.

larsrh:
@S11001001 Would you be inclined to run this through `tut`? Recently, this blog gained the ability to do so automatically. [See this post](https://github.com/typelevel/typelevel.github.com/blob/development/posts/2016-10-17-minicheck.md) for an example.

edmundnoble:
So awesome. I remember seeing something similar on Robert Harper's blog.

milessabin:
This is great, but can't we get there a lot more quickly?

Each class has a unique, finite, name (including path) which corresponds to a finite character string. There are countably many finite strings so, by correspondence, there are countably many classes, ie. ω.

Now consider the set of unary function types A => B for all classes A and B. The size of this set is ω^ω, which is bigger than ω by the usual Cantorian diagonal argument.

QED.

edmundnoble:
@milessabin not sure if you're trolling, but surely you're being impredicative. Consider the set of unary function *classes* for all classes A and B. The size of this set is also omega^omega. You then arrive at a contradiction because the set of classes contains a set larger than itself.

milessabin:
Not trolling. There isn't a set of unary function _classes_. There can't be, because all classes have names and there aren't enough names to go around.

edmundnoble:
If there's only one unary function class and an infinite number of unary function types, does that not get you to your conclusion immediately?

S11001001:
> So awesome. I remember seeing something similar on Robert Harper's blog.

@edmundnoble Yep! I've deliberately borrowed his terminology here, too. I wanted to target a different audience and work in more typeish goodies along the way.

> This is great, but can't we get there a lot more quickly?

@milessabin Totally. This is the scenic route, with lots more good photo opportunities.

posts/2017-02-13-more-types-than-classes.md
S11001001: I do! Thanks.

larsrh:
> Totally. This is the scenic route, with lots more good photo opportunities.

I agree with that sentiment.

posts/2017-02-13-more-types-than-classes.md
S11001001: It does hurt if you make this confusion, but it'd kill the narrative to address it, and I've already included a lot more about the semantics of `val` vs `var` than most articles that tackle just those.

Anyway there's always [another Robert Harper post](https://existentialtype.wordpress.com/2012/02/01/words-matter/) to address that confusion :)

alexandru:
One thing I’ve noticed is that when I, as a reader, observe an incomplete statement, I tend to skip over the rest of the article in my dire need to provide feedback to the author, letting the world know that he’s wrong :-)

Classes are relevant for OOP languages and are (also a) runtime thing because the OOP (subtype) polymorphism works at runtime, for any call-site the JVM needing to know (1) the interface + method you try to call and (2) the object’s vtable in order to link that method with an actual implementation, according with Java's single dispatching rules.

Such a reader, familiar with what OOP means and Java, will nod starting from the first section where it is said more or less that types are about variables and not objects, which isn't true, unless you also think dynamic languages have a single type, which IMO is a subjective opinion and does not help in making a distinction at all. A more useful comparison is to talk about what gets checked and inferred at compile time and what gets deferred to runtime, which is like a slider that gets moved closer to compile time for languages like Haskell and ML and closer to runtime for languages like Lisp.

> That is why ... total type erasure is a desirable feature rather than a design flaw—plugging these leaks gives the programmer as much freedom to abstract by information-hiding as she wishes. 

I've already heard the opinion from @travisbrown that he'd prefer total type erasure, but I don't agree. Even C++ has [RTTI](https://en.wikipedia.org/wiki/Run-time_type_information), allowing it to safely perform casts, for an OOP language you need casting and if you have casting you need `isInstanceOf` checks as well, otherwise you can have memory vulnerabilities. If you think `isInstanceOf` checks are not useful in the presence of type erasure for generics, I personally disagree, because those checks are going to happen anyway, done by the runtime and as a developer I'd rather have access to such checks, especially for doing optimizations.

posts/2017-02-13-more-types-than-classes.md
```diff
@@ -0,0 +1,702 @@
...
+```scala
+scala> (greeting: String, greeting: java.io.Serializable,
+        greeting: CharSequence, greeting: Comparable[String],
+        greeting: AnyRef)
+res3: (String, java.io.Serializable, CharSequence,
+       Comparable[String], AnyRef) =
+  (hi there!,hi there!,hi there!,hi there!,hi there!)
+```
+
+So we have exhausted the classes, but aren’t quite done with types.
```
propensive: Could you also include (either here, or later) a mention of the singleton literal type, as distinct from the value type? Whilst I don't think it adds all that much, "we...aren't quite done with types" suggests that we had only missed one type. It might then be necessary to note that the singleton literal types arising from two different (but equal) values still conform, even though their value types don't... (You may equally decide, quite justifiably, that the subtleties of Scala's inference and type-widening are too complicated for this article, and skip them!)

S11001001:
> Would you be inclined to run this through tut?

@larsrh I'll play with it this week, but would rather publish today with what exists now.

> misunderstandings about classes vs types, and also type conformance vs type equality. (You touch on the latter point, though I think there's probably enough material for another blog article on this.)

@propensive I think comparing `Leibniz` and `Liskov` is the best approach to understanding the difference. Alas, such a treatment also requires delving further into introductory variance, a well which has by now run dry for me. Perhaps one of Typelevel's many fans of subtyping might like to have a go?

> Given the title, I might also be tempted to mention, as an aside, the existence of types which don't (and can't possibly) have values, but which the compiler can still talk about

@propensive [shh :)](https://usecanvas.com/typetype/empty-types-and-nonsense-types/49xTrTpC0zRP2bzFvpdEgm)

posts/2017-02-13-more-types-than-classes.md
S11001001: @propensive I've touched on this briefly in 2bc3166, as it happens to fit quite well with the existing text. Thanks!

S11001001:
> unless you also think dynamic languages have a single type

@alexandru Yep! That's me. I mention this later in the article.

larsrh:
> I'll play with it this week, but would rather publish today with what exists now.

Fine with me :+1:

S11001001:
@tpolecat @non @larsrh @edmundnoble @milessabin @alexandru Thank you very much for your comments on the draft.

posts/2017-02-13-more-types-than-classes.md
S11001001: Almost forgot :) 1f53f2c

PR #145, opened 2017-02-27
Tutify "There are more types than classes"

S11001001:
As suggested in #143 by @larsrh.  I've only guarded the blocks that would output something close enough to the original manual blocks, and also have all dependency blocks evaluated.

/cc @tpolecat


larsrh:
Great, thanks!

PR #146, opened 2017-02-27
New post: "Four ways to escape a cake"

S11001001:
For publication on March 1.

(There is a [tut patch](https://github.com/S11001001/typelevel.github.com/compare/four-ways-to-escape-a-cake...S11001001:four-ways-to-escape-a-cake+tut) to this, similar in spirit to #145; lacking output, it is more difficult to review.)

### DIFF
--- a/posts/2017-03-01-four-ways-to-escape-a-cake.md
+++ b/posts/2017-03-01-four-ways-to-escape-a-cake.md
@@ -13,7 +13,8 @@ within traits, as seen in `scala.reflect.Universe`, ScalaTest, and
 other Scala styles, seems to be infectious. By that, I mean once you
 define something in a trait to be mixed in, to produce another
 reusable module that calls that thing, you must define *another*
-trait, and so must the downstream of that, and so on and so forth.
+trait, and so must anyone using *your* module, and so on and so forth.
+You effectively become “trapped in the cake”.
 
 However, we can use type parameters that represent **singleton types**
 to write functions that are polymorphic over these “cakes”, without
--- a/posts/2017-03-01-four-ways-to-escape-a-cake.md
+++ b/posts/2017-03-01-four-ways-to-escape-a-cake.md
@@ -459,11 +459,120 @@ by declaring the universe! All of the non-redundant information is
 represented in the singleton type index. So even where the above
 type-refinement technique works (and it does in many cases), it’s
 *still* redeclaring things that ought to be derivable from the “mere”
-fact that `U` is a singleton type. If at all possible, I would like to
-find a way to use that information, and only that information, to make
-a complete escape from the cake.
+fact that `U` is a singleton type.
 
-As it stands, we’re almost out.
+## The fact that it’s a singleton type
+
+*(The following is based on enlightening commentary by Daniel Urban on
+an earlier draft.)*
+
+Let’s examine the underlying error in `stepTwiceEx` more directly.
+
+```scala
+scala> def fetchIter[U <: LittleUniverse](
+    h: U#Haystack): U#Needle => U#Needle = h.iter
+<console>:14: error: type mismatch;
+ found   : _1.type(in method fetchIter)#Needle
+             where type _1.type(in method fetchIter) <: U with Singleton
+ required: _1.type(in value $anonfun)#Needle
+             where type _1.type(in value $anonfun) <: U with Singleton
+           h: U#Haystack): U#Needle => U#Needle = h.iter
+                                                    ^
+```
+
+It’s a good thing that this doesn’t compile. If it did, we could do
+
+```scala
+fetchIter[LittleUniverse](lu.haystack)(anotherU.haystack.init)
+```
+
+Which is unsound.
+
+[§3.2.1 “Singleton Types”](http://www.scala-lang.org/files/archive/spec/2.12/03-types.html#singleton-types) of
+the specification mentions this `Singleton`, which is in a way related
+to singleton types.
+
+> A *stable type* is either a singleton type or a type which is
+> declared to be a subtype of trait `scala.Singleton`.
+
+Adding `with Singleton` to the upper bound on `U` causes `fetchIter`
+to compile! This is sound, because we are protected from the above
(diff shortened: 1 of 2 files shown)


posts/2017-03-01-four-ways-to-escape-a-cake.md
```diff
@@ -0,0 +1,476 @@
...
+them.
+
+## Extracting more types
+
+We have the same problem with the function version.
+
+```scala
+def stepTwiceHaystack[U <: LittleUniverse](
+    h: U#Haystack, n: U#Needle): U#Needle =
+  h.iter(h.iter(n))
```
durban: Note, that it seems that we can make this work (at least in this specific case) by using `Singleton`:

```scala
def stepTwiceHaystack[U <: LittleUniverse with Singleton](
  h: U#Haystack, n: U#Needle
): U#Needle = h.iter(h.iter(n))

stepTwiceHaystack[lu.type](lu.haystack, lu.haystack.init) // works fine
```

And we can use this in turn to implement the extension method:

```scala
class Syntax[U <: LittleUniverse with Singleton](u: U) {
  def stepTwice(n: U#Needle): U#Needle =
    stepTwiceHaystack[U](u.haystack, u.haystack.init)
}

implicit def Syntax[U <: LittleUniverse](u: U): Syntax[u.type] =
  new Syntax[u.type](u)

lu stepTwice lu.haystack.init // also seems to work fine
```

I don't pretend to completely understand how/why this works, but it seems to me, that (at least in some cases) we _can_ constrain `U` to be a singeton type...

S11001001: Cool! (I have tried `Singleton` in other cases, but couldn't get anywhere, so glad to see this.) I'll test it out later, and if all seems well rewrite to declare that all four escape paths are open :+1:

milessabin: It might also be worth your while playing with TLS with `-Yliteral-types` it has fixes for a couple of singleton type related bugs. Most likely the fix for [SI-5103](https://issues.scala-lang.org/browse/SI-5103) is relevant.

S11001001: @durban I've added a bunch about this in 13ecc3f. Thanks again.

@milessabin Indeed, 5103 seems to improve inference. Neat.

```scala
scala> def stepTwiceSingly[U <: LittleUniverse with Singleton](
    h: U#Haystack, n: U#Needle): U#Needle = {
  h.iter(h.iter(n))
}

scala> stepTwiceSingly(lu.haystack, lu.haystack.init)
res1: lu.Needle = Needle()
```

However, TLS 2.12.1 `-Yliteral-types` seems to have the same difference in treatment between existential and non-existential vals described in the new section.

milessabin:
I think you're making a bit of a meal of this. Rather than trying to pull everything out of the cake I would "extend" the cake and push the operations in, right up until the end.

```scala
object CakeDemo {
  final case class LittleUniverse() {
    val haystack: Haystack = Haystack()

    final case class Haystack() {
      def init: Needle = Needle()
      def iter(n: Needle): Needle = n
    }

    final case class Needle()
  }

  class WithUniverse[U <: LittleUniverse](val u: U) {
    import u._

    def stepTwice(n: Needle): Needle =
      haystack.iter(haystack.iter(n))      

    def run: Needle =
      stepTwice(haystack.init)
  }

  object WithUniverse {
    // Workaround for SI-5103
    def apply[U <: LittleUniverse](u: U): WithUniverse[u.type] = new WithUniverse(u)
  }

  final val lu: LittleUniverse = LittleUniverse()

  val res: lu.Needle = WithUniverse(lu).run
}
```

I also see the occurrence of a type projection as a type smell ... if you use a '#' you are almost always asking for trouble.

posts/2017-03-01-four-ways-to-escape-a-cake.md
durban: :+1:

S11001001:
@milessabin Since there are many universe-polymorphic functions that do not actually need the universe to operate, just its type-level reflection, my goal was to find ways to do that. This oft-unused universe argument needing to be passed everywhere looks to me as an albatross on the signatures of those pieces of code. I want to escape cakes, not get mired in the outer layers :)

I've also written about the dangers of type projection [in a prior post](http://typelevel.org/blog/2015/07/23/type-projection.html).

milessabin:
@S11001001 perhaps so, but I suspect that any "pulling out" solution can be rewritten more simply in  "pushing in" form. Your mission, if you choose to accept it, is to find a counterexample to this claim.

S11001001:
@milessabin Sure, I'll accept that challenge.

```scala
// library.scala
import scala.reflect.api.Types

object FancyLibrary {
  def monoIncreasing[U <: Types with Singleton](xs: List[U#Type]): Boolean =
    xs match {
      case fst :: snd :: tl =>
        fst <:< snd && monoIncreasing(snd :: tl)
      case _ => true
    }
}

// app.scala
import FancyLibrary._
import scala.reflect.runtime.universe.typeOf

object Blah {
  def one = monoIncreasing(List(typeOf[String],
    typeOf[CharSequence], typeOf[AnyRef]))
  // evals to true

  def two = monoIncreasing(List(typeOf[String], typeOf[Int]))
  // evals to false
}

// tested with TLS 2.12.1, -Yliteral-types
```

Setting that aside, there are two things I'm more interested in.

First: with some regularity, I say things like "cakes create hidden type-level complexity". Whether or not I define an abstraction over it, it remains a property of the cake itself, not `fetchIter` et al, that the expression `h.iter` has a type quantified by a potentially existential value, and this is belied by the surface syntactic simplicity. One of the goals of this article is to show what lies beneath.

Second: I wish to write fully generalized types for my definitions, for the sake of being honest with their callers. If my library function definition demands an argument that it doesn't use, it isn't being completely honest. If Scala makes that ideal inconvenient to live up to, well, better that than if Scala made it impossible.

PR #148, opened 2017-03-04
make rss great again

jarrodwb:
I am a user of rss feed aggregators and find the current rss feed not as
useful as it could be. You can checkout the current state by looking at
http://typelevel.org/blog/feed.rss on feedbucket.com. The list of blog posts is
limited to 10 entries and the descriptions are very terse.

The terseness, I think, is a bug with the `{% if post.excerpt %}` line.
At some point Jekyll decided to auto-generate excerpts if one is not
given in the YAML header. So I think this `if` statement is always true.

I greped the posts directory and found no use of the `excerpt` variable
so I made the decision to just remove the use of the variable in the
feed.rss file.

The end result is a useful rss feed. It can be previewed at
feedbucket.com using this link
https://gist.githubusercontent.com/jarrodwb/003ea9995134bde814e6d3545529e483/raw/bbc5d244c734f55fce77dc5b08fc8140be32abea/rss-fixes.rss.

The formatting is beautiful on feedbin.com. This is the aggregator I use.

### DIFF
--- a/_includes/license.html
+++ b/_includes/license.html
@@ -1,0 +1,4 @@
+<h2>Licensing</h2>
+<p>
+  Unless otherwise noted, all content is licensed under a <a href="https://creativecommons.org/licenses/by/3.0/deed.en_US">Creative Commons Attribution 3.0 Unported License</a>.
+</p>
--- a/_layouts/post.html
+++ b/_layouts/post.html
@@ -7,6 +7,8 @@ layout: default
 
   {{ content }}
 
+{% include license.html %}
+ 
   <div class="post__author js-fade-in">
     {% assign author = site.data.authors[page.meta.author] %}
     {% if author %}
--- a/about.html
+++ b/about.html
@@ -45,10 +45,7 @@ title: About
     {{ conduct | markdownify }}
   </div>
 
-  <h2>Licensing</h2>
-  <p>
-    Unless otherwise noted, all content is licensed under a <a href="https://creativecommons.org/licenses/by/3.0/deed.en_US">Creative Commons Attribution 3.0 Unported License</a>.
-  </p>
+{% include license.html %}
 
   <h3>Images</h3>
   <p>
--- a/blog/feed.rss
+++ b/blog/feed.rss
@@ -14,7 +14,7 @@ layout: empty
         {% if post.meta.author %}
           <dc:creator>{{ post.meta.author | xml_escape }}</dc:creator>
         {% endif %}        
-        <description>{{ post.content | xml_escape }}</description>
+        <description><![CDATA[{{ post.content }}{% include license.html %}]]></description>
         <pubDate>{{ post.date | date_to_rfc822 }}</pubDate>
         <link>{{ site.baseurl }}{{ post.url }}</link>
         <guid isPermaLink="true">{{ site.baseurl }}{{ post.url }}</guid>
--- a/blog/feed.rss
+++ b/blog/feed.rss
@@ -10,8 +10,8 @@ layout: empty
     {% for post in site.posts %}
       <item>
         <title>{{ post.title | xml_escape }}</title>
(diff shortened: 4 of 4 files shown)


jarrodwb:
Hi. :-) Thanks for the review.

The notes on syntax highlighting and MathJax support are the responsibility of the RSS reader. Feedbin for instance provides syntax highlighting and code blocks for the new RSS feed where as FeedBucket does not. MathJax is similar but I am not sure if Feedbin supports it. Which article contains MathJax? I will take look and see if Feedbin does the right thing.

I agree on the the other note on copyright. I can put the same notice as is on http://typelevel.org/about.html which states all content is under CCA3.0 unless otherwise noted.

We should also definitely add the author. I missed that.

Cheers. :-)

jarrodwb:
I am not sure if I have the best solution to the license issue but I put the CC BY 3.0 license into the RSS feed for the whole channel. It is up to the RSS readers/aggregators to display this. I also added the license to each post by changing the  post layout. This does not show up in the page content variable so it is not propagated to the RSS items description.

I want to do something like this `{% include license.html | xml_escape %}` in the feed.rss file but I get a liquid parsing error.

I think what I provided in my last commit is sufficient but is not as nice as I would like.

jarrodwb:
I used an CDATA tag to fix the problem I mentioned before. Now all items in the RSS feed have a license at the end of the article.

S11001001:
Hi @jarrodwb, thanks for the update to incorporate copyright+license information; looks nice.

> The notes on syntax highlighting and MathJax support are the responsibility of the RSS reader.

I don't think this is true, especially when we can render something more likely to succeed for users. (By contrast, we can't do anything about a reader ignoring `dc:creator`, but we've done everything we can in that area.)

> Feedbin for instance provides syntax highlighting and code blocks for the new RSS feed where as FeedBucket does not.

Would you mind sharing a screenshot of Feedbin's rendering of, say, the code block beginning with `def twoInitsFromAHaystack` from ["Four ways to escape a cake"](http://typelevel.org/blog/2017/03/01/four-ways-to-escape-a-cake.html#getting-two-needles-without-a-universe)?

> Which article contains MathJax? I will take look and see if Feedbin does the right thing.

One example is ["Symbolic operators and type classes for Cats"](http://typelevel.org/blog/2015/08/07/symbolic-operators.html):

We call an operation ⊕ associative, if for all a, b and c, a⊕(b⊕c)=(a⊕b)⊕c holds.

jarrodwb:
I took some screenshots. It looks like even Feedbin does not automatically render MathJax. :-/

I don't have any ideas about how to render the posts before they hit the RSS feed. Maybe we can use what we have and improve on it later. If the rendering is a real blocker for this PR than maybe we can cherry pick some of my changes and abandon the idea of having complete posts shared through the RSS feed. Another option is to open some issues and simply close this PR.

https://drive.google.com/open?id=0B1ygQn_r10X_Z2tSNE5NbWw5T3M
https://drive.google.com/open?id=0B1ygQn_r10X_OTZWSFhRQlpTOHc

S11001001:
> I took some screenshots. It looks like even Feedbin does not automatically render MathJax. :-/

It also does not handle syntax highlighting correctly. Note how everything after the `#` on the second line is treated as a comment; also, `def` is keyword-highlighted, but not `val`; hmm, Python, maybe?

It's worth spinning up two issues for these: #149, #150. Maybe @larsrh has some opinions on the rest.

larsrh:
I'm happy to merge this as it stands. As far as I can tell it is a strict improvement over the status quo. Is that right?

jarrodwb:
Is there any chance of getting a verdict on this? I would like to cross it off my todo list. :-)

larsrh:
Sorry, I missed that @S11001001 updated his review. Merging now.

@jarrodwb, thanks for the contribution!

PR #151, opened 2017-03-19
guest post by Tomas Mikula: "Equivalence versus Equality"

larsrh:
This has been edited until the point where the `hasheq` library is needed. I still need to contemplate whether to include some toy version of the actual code so that we can run it through tut, or to leave it as it is and link to the repository.


S11001001:
@larsrh I'd personally wait until a release is available to fully tut-ify, so tut `dependencies` can be used.  But I'm still pretty conservative when it comes to `tut` usage.

larsrh:
@S11001001 The problem is that `hasheq` hasn't been released yet. But I'm starting to come to the conclusion that including a toy version would require way too much code, which would distract from the message.

larsrh:
I added the commit hash of the current version of hasheq at the bottom of the post. I say we can merge this now.

TomasMikula:
I made a release: https://oss.sonatype.org/content/repositories/releases/com/github/tomasmikula/

Should appear on Maven Central shortly.

larsrh:
tutified in f68ea44

PR #159, opened 2017-03-25
Add Monix as a project with full membership

alexandru:
Related to: https://github.com/typelevel/general/issues/61

This adds the description for Monix in `_projects` and removes it from the incubator list, not sure if I have to do anything else.


larsrh:
That's it, thanks!

PR #162, opened 2017-04-13
add fragment IDs for each project to support deep links

bfritz:
The project page is fairly long This would allow deep links like http://typelevel.org/projects/#scala-exercises , for example in README badges.

`slugify: 'raw'` converts to lowercase and replaces spaces, not allowed in HTML IDs, with dashes.

### DIFF
--- a/projects/index.html
+++ b/projects/index.html
@@ -14,7 +14,7 @@ class: projects-page
       <div class="projects__project expandable-core">
         <div class="projects__project-header">
           <a href="{{ project.github }}">
-            <h4>
+            <h4 id="{{ project.title | slugify: 'raw' }}">
               <span class="ico-repo"></span>
               {{ project.title }}
             </h4>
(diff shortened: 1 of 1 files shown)


ceedubs:
I like this.

Do you think that we should have any sort of prefixes on the IDs to avoid a potential element id collision?

bfritz:
@ceedubs Thanks for taking a look.  Do you have a specific collision scenario in mind?

IMO http://typelevel.org/projects/#cats is nicer than http://typelevel.org/project/#xyz-cats for external links to the projects page.

Trying to think through cases where a prefix is useful...
* Two projects with same name--modulo spaces and case.  E.g. "ScalaExercises" and "scala exercises".  Seems unlikely since names so similar will be confusing in general.
* Later on an ID is added, say `id="cats"`, for some other reason.  In that case, it seems like the ID added later might be better to prefix.  The ones in the PR are already positioned well for linking to from external sites.
* The project page is restructured so there are two or more sections about each project.  It would probably make sense to pick the "meatiest" of the sections and keep the ID on that one.

Let me know if you had some other case in mind.  I'm happy to add a prefix if you and the other maintainers think it's a good idea.

larsrh:
I think @bfritz' change is good. If we want to add some more anchors at a later time we can prefix _those_.

larsrh:
Thanks!

ceedubs:
@bfritz I was just being paranoid about a case in which there's a library with a name that might happen to conflict with a reasonable HTML element ID. I'm sure it will be fine though :)

bfritz:
@ceedubs Gotcha.  Could definitely happen but seems unlikely.  Integrating something like [html-proofer](https://github.com/gjtorikian/html-proofer) _might_ be a way to catch it during CI.

larsrh:
@bfritz Sounds like an excellent idea. Another pull request, maybe? :smile:

PR #169, opened 2017-04-30
New post: An IO monad for cats

larsrh:
Review by @djspiewak


posts/2017-05-02-io-monad-for-cats.md
```diff
@@ -0,0 +1,404 @@
...
+
+What we need is a data type which allows us to encapsulate Scala-style side-effects in the form of a *pure* value, on which referential transparency holds and which we can compose using other well-defined abstractions, such as `Monad`.  Scalaz defines two such data types which meet these criteria: `scalaz.effect.IO` and `scalaz.concurrent.Task`.  But in practice, nearly everyone uses `Task` instead of `IO` because of its support for *asynchronous* effects.
+
+Cats does not define any such abstraction, and what's worse is the cats *ecosystem* also doesn't really provide any such abstraction.  There are two `Task` implementations that are relatively commonly used with cats – namely, `monix.Task` and `fs2.Task` – but these are not part of cats per se, nor are they deeply integrated into its abstraction hierarchy.  Additionally, the proliferation of broadly equivalent options has led to confusion in the ecosystem, as well as some justifiable fear, uncertainty and doubt on the part of those evaluating cats as a potential dependency.
```
alexandru: I'm not sure about this phrasing:

> the proliferation of broadly equivalent options has led to confusion in the ecosystem, as well as some justifiable fear, uncertainty and doubt on the part of those evaluating cats as a potential dependency

I haven't heard this opinion often and this phrasing here has the potential to create the FUD it is talking about.

I would instead say something like this: 

Additionally, the proliferation of options can be problematic for beginners that don't know what to choose, a problem known as [the Paradox of Choice](https://en.wikipedia.org/wiki/The_Paradox_of_Choice).

posts/2017-05-02-io-monad-for-cats.md
```diff
@@ -0,0 +1,404 @@
...
+
+## Introducing cats-effect
+
+The [cats-effect](https://github.com/typelevel/cats-effect) project aims to change all of that.  The goal of cats-effect is to provide an "easy default" `IO` type for the cats ecosystem, deeply integrated with cats-core, with all of the features and performance that are required for real world production use.  Additionally, cats-effect defines a set of abstractions in the form of several typeclasses which describe what it means to *be* a pure effect type.  These abstractions are extremely useful both in enabling MTL-style program composition and to ensure that other pre-existing `Task` implementations remain first-class citizens of the ecosystem.  `IO` does not overshadow `monix.Task` or `fs2.Task`; it *complements* them by providing a set of abstractions and laws which allow users to write safe, parametric code which supports each of them equally.
```
alexandru: `monix.Task` -> `monix.eval.Task` 

(if the intention is to give the full path, like in `scalaz.concurrent.Task`, instead of `scalaz.Task`)

posts/2017-05-02-io-monad-for-cats.md
```diff
@@ -0,0 +1,404 @@
...
+
+The above will run `program` *twice*.  So clearly, referential transparency is out the window whenever we do this, and we cannot expect the normal laws and abstractions to remain sound in the presence of this function.
+
+### A sidebar on `Future`'s eager evaluation
+
+As Viktor Klang is fond of pointing out, `Future` doesn't *need* to evaluate eagerly.  It is possible to define an `ExecutionContext` in which `Future` defers its evaluation until some indefinitely later point.  However, this is not the default mode of operation for 99% of all `Future`s ever constructed; most people just use `ExecutionContext.global` and leave it at that.  Additionally, if someone hands me an arbitrary `Future`, perhaps as a return value from a function, I really have no idea whether or not that `Future` is secretly running without my consent.  In other words, the referential transparency (or lack thereof) of functions that I write using `Future` is dependent on the runtime configuration of some other function which is hidden from me.  That's not referential transparency anymore.  Because we cannot be *certain* that `Future` is deferring its evaluation, we must defensively assume that it is not.
```
alexandru: 👍 it's good to be precise

posts/2017-05-02-io-monad-for-cats.md
```diff
@@ -0,0 +1,404 @@
...
+
+This, in a nutshell, is precisely why you should only use `Future` when interacting with legacy code or third-party APIs which require it, and when you do use `Future`, you should treat it as a side-effect just like `println` or `launchTheMissiles`.
```
alexandru: `Future` is an integral part of Scala's ecosystem and there are contexts in which `Future` is actually useful. When working in an imperative / dirty context, sometimes you really need `Promise` / `Future` references.

To give an example, in Monix I'm building a [AsyncSemaphore](https://github.com/monix/monix/blob/master/monix-execution/shared/src/main/scala/monix/execution/misc/AsyncSemaphore.scala) that uses Futures internally, because that was the best approach and then I'm wrapping that into a [TaskSemaphore](https://github.com/monix/monix/blob/master/monix-eval/shared/src/main/scala/monix/eval/TaskSemaphore.scala).

I would say this about `Future`: it's a low-level concurrency primitive and people should treat it as a low level concurrency primitive.

posts/2017-05-02-io-monad-for-cats.md
```diff
@@ -0,0 +1,404 @@
...
+
+## Conclusion
+
+The lack of a production-ready `Task`-like type fully integrated into the cats ecosystem has been a sticking point for a lot of people considering adopting cats.  With the introduction of [cats-effect](https://github.com/typelevel/cats-effect), this should no longer be a problem!  As of right now, the only releases are snapshots with hash-based versions, the latest of which can be found in the maven badge at the top of the readme.  These snapshots are stable versions (in the repeatable-build sense), but they should not be considered stable, production-ready, future-proof software.  We are quickly moving towards a final 0.1 release, which will depend on cats-core and will represent the stable, finalized API.
```
alexandru: Is it "cats" or is it "Cats"? On http://typelevel.org/cats/ I see that the name is capitalized.

alexandru: We should really have a first 0.1 version before releasing this article IMO.

The hash-based versioning is cool, but it's an alternative to snapshots and people won't want to depend on that.

larsrh: For `tut` I've used the hash-based release, which I think is "good enough" for a blog post.

posts/2017-05-02-io-monad-for-cats.md
larsrh: I'll let @djspiewak decide on this one.

posts/2017-05-02-io-monad-for-cats.md
larsrh: Sounds good to me, but it's @djspiewak's call.

djspiewak: I actually don't agree with @alexandru here. When I want a concurrency primitively, Future is far too high level and far too slow. Something like fs2's Actor is generally preferable when I'm in those sorts of situations.  Also, avoiding the uncontrolled memorization is another reason not to use Future for that sort of thing.

I'm glad you've been able to put Future to good use, but I just haven't found the same, and other primitives have always met my needs better than Future could.

posts/2017-05-02-io-monad-for-cats.md
djspiewak: I've had numerous people independently express the problem I stated here.  It's definitely a real issue.

djspiewak: Perhaps the phrasing could be softened though. As you said, it does have the potential to fan some flames.

larsrh:
@djspiewak Do you want to update some wording or shall we go ahead and merge?

djspiewak:
@larsrh Going to make a minor update to some wording. Will post here within an hour or so.

tpolecat:
hang on pls, i have a few comments

posts/2017-05-02-io-monad-for-cats.md
```diff
@@ -0,0 +1,404 @@
...
+
+
+Haskell is a pure language.  Every Haskell expression is *referentially transparent*, meaning that you can substitute that expression with its evaluated result without changing the program.  Or, put into code:
+
+```haskell
+f e e
+-- is equivalent to
+let x = e in f x x
+```
+
```
tpolecat: People get angry when we assume they understand Haskell. Can we make this a bit more clear, like

```haskell
-- this program
f expr expr -- apply function f to arguments expr, expr

-- is equivalent to this one, which factors out `expr`
let
  x = expr -- introduce a new binding `x`
in
  f x x
```

posts/2017-05-02-io-monad-for-cats.md
```diff
@@ -0,0 +1,404 @@
...
+program
+
+// must be the same as!
+
+program
+```
+
+If this weren't the case, then we would be in trouble when trying to construct examples like the Haskell one from earlier.  But there is an implication here that is quite profound: `IO` cannot eagerly evaluate its effects, and similarly cannot memoize its results!  If `IO` were to eagerly evaluate or to memoize, then we could no longer replace references to the expression with the expression itself, since that would result in a *different* `IO` instance to be evaluated separately.
+
+This is precisely why `scala.concurrent.Future` is *not* a suitable type for encapsulating effects in this way: it isn't referentially transparent!  `Future` evaluates eagerly (sort of, see below) and memoizes its results, meaning that a `println` inside of a given `Future` will only evaluate *once*, even if the `Future` is sequenced multiple times.  This in turn means that `val x = Future(...); x; x` is not the same as `Future(...); Future(...)`, which is the very definition of a violation of referential transparency.
```
tpolecat: > This is precisely why `scala.concurrent.Future` is *not* a suitable type for encapsulating effects in this way: it isn't referentially transparent!

I think we need to be more precise here, since Future is a type and RT is a property of terms. And people like to jump on us when we say this kind of thing. Maybe something like

> This is precisely why `scala.concurrent.Future` is *not* a suitable type for encapsulating effects in this way: constructing a `Future` that will [eventually] side-effect is itself a side-effect!

I would also suggest `val x = Future(...); f(x, x)` as the example because `x; x` uses statements, which don't really have a meaning in FP.

posts/2017-05-02-io-monad-for-cats.md
```diff
@@ -0,0 +1,404 @@
...
+The reason it isn't equivalent comes from the different sorts of expressions that we could find in `e`.  For example, what if `e` is `println("hi!")`.  If we make that substitution, our snippet looks like the following:
+
+```scala
+f(println("hi"), println("hi"))
+// isn't really equivalent to!
+val x = println("hi")
+f(x, x)
+```
+
+Clearly these are not the same two programs.  The first prints `"hi"` twice, while the second only prints it once.  This is a violation of referential transparency, and it's why we sometimes say that Scala is an *impure* language.
```
tpolecat: I would add here.

> An expression that is *not* referentially transparent is called a **side-effect**.

posts/2017-05-02-io-monad-for-cats.md
tpolecat: Although I agree that Future is terrible that's not a debate we need have here. I think we can confidently say that it's not appropriate for functional programming and leave it at that.

posts/2017-05-02-io-monad-for-cats.md
```diff
@@ -0,0 +1,404 @@
...
+```tut:silent
+val program = IO {
+  println("Welcome to Scala!  What's your name?")
+  val name = Console.readLine
+  println(s"Well hello, $name!")
+}
+```
+
+But this gives us less flexibility for composition.  Remember that even though `program` is a pure and referentially transparent value, its *definition* is not.  Anything inside the `IO {` … `}` block is not referentially transparent, and so should be treated with extreme care and suspicion.  The less of our program we have inside these blocks, the better!
+
```
tpolecat: Just to hammer this home I would show what it means in terms of substitution. Something like

> … *definition* is not, which is to say, `IO(expr)` is **not** the same as `{ val x = expr; IO(x) }`. Anything …

posts/2017-05-02-io-monad-for-cats.md
djspiewak: That's not actually true though.  An expression which is not referentially transparent *contains* side-effects.

djspiewak:
@tpolecat I couldn't reply to your last point, for some reason…

> Although I agree that Future is terrible that's not a debate we need have here. I think we can confidently say that it's not appropriate for functional programming and leave it at that.

djspiewak:
@larsrh Here's a revised gist.  Didn't feel like forking your PR to submit an actual diff.  :-)  https://gist.github.com/52ed6f8356dd073f6a0080713043175f

posts/2017-05-02-io-monad-for-cats.md
tpolecat: It *is* a side-effect. That's the definition. It's a syntactic property. Fight me.

djspiewak: From the definition (Wikipedia; emphasis mine):

>  In computer science, a function or expression is said to **have** a side effect if it modifies some state outside its scope

Bring it on, Foreman.

djspiewak: I guess the distinction that I'm drawing here is that a side-effect is not "a thing" in the sense that you can grab it and abstract over it.  It's an ephemeral property of an expression, and that *expression* is a thing.  Though I suppose that if I'm drawing that distinction, I should be more precise about what `IO.apply` does: captures expressions which contain side-effects.

tpolecat:
> The problem is that a very common response to this problem domain is "Why not just use Future?" It's important to understand why Future is a poor fit for this problem, as opposed to just the bespoke "it's bad, I promise.

Yes, agree, I'm was just quibbling over the scope of the closing paragraph.

> This, in a nutshell, is precisely why `Future` is not appropriate for functional programming …

Might be easier to defend than a prescription for when you *should* use it.

> This, in a nutshell, is precisely why you should only use `Future` when …

But whatever you go with is fine here. Not a big quibble.

posts/2017-05-02-io-monad-for-cats.md
tpolecat: Bah, what does Wikipedia know?

> I should be more precise about what IO.apply does: captures expressions which contain side-effects.

I think you get this point across pretty well, but yes, the louder the better. This is the operation `Future` doesn't have.

djspiewak:
> Yes, agree, I'm was just quibbling over the scope of the closing paragraph.

Oh, that's fair.  I'll trim that back.  @alexandru also objected to that paragraph.

alexandru:
👍

larsrh:
@djspiewak Are we good to go?

djspiewak:
Pull ze lever!

larsrh:
Levers pulled, shots fired

PR #173, opened 2017-05-22
Remove 'delta' from incubators.

stacycurl:


stacycurl:
Delta is a very slow boil project, I basically commit to it when some problem crops up at work & that's usually about diffing json.

I like the autonomy of experimenting however I like (albeit infrequently), I don't think that makes for a very good open-source project though.

My involvement with typelevel is very sparse too, practically just a 'Sup ?' when passing by other members.

milessabin:
We're sorry to see you go, and would like to assure you that the door is open for you to return in future should you choose to.

PR #187, opened 2017-08-22
Add decline to incubator.yaml

bkirwi:
Rather-belatedly making the change for https://github.com/typelevel/general/issues/80!


larsrh:
No problem!

PR #201, opened 2018-02-16
Sponsor logos for Berlin Summit 2018

davegurnell:
I've also snuck in some tweaks to the way logos are displayed on sponsor pages (see the last commit).


larsrh:
Something's wrong with the Criteo logo – it doesn't render properly on my machine.

davegurnell:
Filling in a chat Lars and I had over email, it looks like this is a problem with Lars' ad blocker. It shouldn't affect most users.

larsrh:
Also, it's fixed now :smile:

PR #202, opened 2018-03-21
Graduate cats-effect

alexandru:
See: https://github.com/typelevel/general/issues/88

NOTE: I don't know if it's best to add it as an extension or as its own project.

It's linked to Cats, so I guess an extension is better.
I hope the permalink works, as I haven't tested it.

I have no preference, it's just that I don't necessarily understand this classification 🙂


larsrh:
> It's linked to Cats, so I guess an extension is better.

LGTM.

alexandru:
@larsrh I think this is the first Cats extension that has a website.

What can we do to link the website instead of, or in addition to the GitHub repository?

alexandru:
Well, I guess people can see the website link when the go to the repository. But a "More" button would have been cool too.

larsrh:
I've changed that in dc7d666. It now links to the website directly.

PR #204, opened 2018-04-14
Rethinking MonadError - new blog post

LukaJCB:
A new blog post about MonadError and its usability. Would love to hear some feedback or proofreading
:)

### DIFF
--- a/posts/2018-04-13-rethinking-monaderror.md
+++ b/posts/2018-04-13-rethinking-monaderror.md
@@ -184,11 +184,11 @@ def raiseErrorStops(e: E, a: A): Boolean =
 In the same way, we want values of `G` to never stop propagating, so `flatMap` should always work:
 
 ```scala
-def flatMapAlwaysWorks(ga: G[A], a: A): Boolean =
-  ga.flatMap(_ => a.pure[G]) =!= ga
+def flatMapAlwaysWorks(ga: G[A], f: A => G[A]): Boolean =
+  ga.flatMap(f) =!= ga
 ```
 
-This law is a bit dodgy as it is very well possible that `ga` contains a valid `A` that just happens to be the same as `a`. Thus we might have to reformulate this one at some point.
+This law is a bit dodgy as it is very well possible that `ga` just happens to be the same as the same as the one returned by `f`. Thus we might have to reformulate this one at some point.
 
 Next we're going to formulate a law that states, that raising an error and then immediatly handling it with a given function should be equivalent to just calling that function on the error value:
 
--- a/posts/2018-04-13-rethinking-monaderror.md
+++ b/posts/2018-04-13-rethinking-monaderror.md
@@ -188,6 +188,8 @@ def flatMapAlwaysWorks(ga: G[A], a: A): Boolean =
   ga.flatMap(_ => a.pure[G]) =!= ga
 ```
 
+This law is a bit dodgy as it is very well possible that `ga` contains a valid `A` that just happens to be the same as `a`. Thus we might have to reformulate this one at some point.
+
 Next we're going to formulate a law that states, that raising an error and then immediatly handling it with a given function should be equivalent to just calling that function on the error value:
 
 ```scala
--- a/posts/2018-04-13-rethinking-monaderror.md
+++ b/posts/2018-04-13-rethinking-monaderror.md
@@ -12,7 +12,7 @@ tut:
   binaryScala: "2.12"
   dependencies:
     - org.scala-lang:scala-library:2.12.4
-    - org.typelevel::cats-effect:1.1.0
+    - org.typelevel::cats-core:1.1.0
 
 ---
 
@@ -41,7 +41,7 @@ Well, it turns out there a few instances where it might be useful, but as we'll 
 
 Now before we continue, let's look at the `MonadError` type class in a bit more detail.
 `MonadError` currently comprises two parts, throwing and catching errors.
-To begin let's have a look at the `throw` part:
+To begin let's have a look at the `throw` part, sometimes also called `MonadThrow`:
 
 ```scala
 trait MonadError[F[_], E] extends Monad[F] {
@@ -52,7 +52,7 @@ trait MonadError[F[_], E] extends Monad[F] {
 ```
 
 This looks fine for now, but one thing that strikes me is that the `F` type seems to "swallow" errors.
-If we look at `F[A]` we have no clue that it might actually yield an error of type `A`, that fact is not required to be represented at all.
+If we look at `F[A]` we have no clue that it might actually yield an error of type `E`, that fact is not required to be represented at all.
 However, that's not a really big issue, so now let's look at the `catch` part:
 
 ```scala
@@ -92,7 +92,7 @@ also returns an `F[A]`. This method takes a pure function `E => A` and thus can 
 For `IO`-like types this is somewhat excusable as something like an unexceptional `IO` is still very uncommon, but for simple data types like `Either` or `Some` that function should just return an `A`, since that's the only thing it can be.
 Just like with `attempt`, we can infinitely chain calls to `handleError`, as it will never change the type.
 
-Ideally our type system should stop us from being able to right this nonsensical code and give us a way to show anyone reading the code that we've already handled errors.
+Ideally our type system should stop us from being able to write this nonsensical code and give us a way to show anyone reading the code that we've already handled errors.
 Now I'm not saying that the functions on `MonadError` aren't useful, but only that they could be more constrained and thus more accurate in their representation. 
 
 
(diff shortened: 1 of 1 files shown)


posts/2018-04-13-rethinking-monaderror.md
```diff
@@ -0,0 +1,456 @@
...
+
+The problem I see with `MonadError` is that it does not address the fundamental difference between these two types of instances. I can pattern match an `Option[A]` with a default value to get back an `A`. With `IO` that is just not possible. So these two groups of types are pretty different, when does it actually make sense to abstract over both of them?
+Well, it turns out there a few instances where it might be useful, but as we'll see later, I'm proposing something that will be equally useful to both groups.
+
+Now before we continue, let's look at the `MonadError` type class in a bit more detail.
+`MonadError` currently comprises two parts, throwing and catching errors.
+To begin let's have a look at the `throw` part:
+
+```scala
+trait MonadError[F[_], E] extends Monad[F] {
```
edmundnoble: Should this be `MonadThrow`?

posts/2018-04-13-rethinking-monaderror.md
```diff
@@ -0,0 +1,456 @@
...
+
+```scala
+def raiseErrorStops(e: E, a: A): Boolean =
+  F.raiseError[A](e).flatMap(_ => a.pure[F]) === F.raiseError[A](e)
+```
+
+In the same way, we want values of `G` to never stop propogating, so `flatMap` should always work:
+
+```scala
+def flatMapAlwaysWorks(ga: G[A], a: A): Boolean =
```
edmundnoble: This law looks wrong to me, `G` in general may have effects.

posts/2018-04-13-rethinking-monaderror.md
```diff
@@ -0,0 +1,456 @@
...
+For example try to implement the above `MonadBlunder[F[_, _], E]` for the standard `cats.effect.IO`.
+It's not going to be simple, whereas with the current encoding we can easily encode both `Either` and `IO`. 
+
+Next we're going to look at laws we can define to make sense of the behaviour we want.
+The first two laws should be fairly obvious. 
+If we `flatMap` over a value created by `raiseError` it shouldn't propogate:
+
+```scala
+def raiseErrorStops(e: E, a: A): Boolean =
+  F.raiseError[A](e).flatMap(_ => a.pure[F]) === F.raiseError[A](e)
```
edmundnoble: This is guaranteed by parametricity. So there's no need to test it.

posts/2018-04-13-rethinking-monaderror.md
```diff
@@ -0,0 +1,456 @@
...
+```
+
+In the same way, we want values of `G` to never stop propogating, so `flatMap` should always work:
```
Blaisorblade: Typo: propagating

posts/2018-04-13-rethinking-monaderror.md
```diff
@@ -0,0 +1,456 @@
...
+```
+
+And this approach works as well, however now we've forced the two type parameter shape onto implementors. This `MonadBlunder` has the following kind `(* -> * -> *) -> * -> *`.
+This means we can very easily define instances for types with two type parameters like `Either`.
+However, one issue might be that it's much easier to fit a type with two type parameters onto a type class that expects a single type constructor `(* -> *)` than to do it the other way around.
+
+For example try to implement the above `MonadBlunder[F[_, _], E]` for the standard `cats.effect.IO`.
+It's not going to be simple, whereas with the current encoding we can easily encode both `Either` and `IO`. 
+
+Next we're going to look at laws we can define to make sense of the behaviour we want.
```
Blaisorblade: For which of the two typeclasses? Did you pick one? Could you say it?

posts/2018-04-13-rethinking-monaderror.md
LukaJCB: Hmmm, I think you're right, I adapted the law from here: https://github.com/purescript/purescript-transformers/blob/master/src/Control/Monad/Error/Class.purs#L18

posts/2018-04-13-rethinking-monaderror.md
LukaJCB: Yeah, that's true, was thinking about that too, it might make more sense to express this law as 
`ga.flatMap(_ => a.pure[G]) =!= ga`, though I'm still not satisfied with that.

posts/2018-04-13-rethinking-monaderror.md
LukaJCB: I.e. you can say that of the function given to `flatMap` can never be called due to there being no value of `A`, that's one side, but this law also requires that the `flatMap` call does absolutely nothing at all, which is a very slightly different statement.
I'm not sure I like it tbh, I'm going to have to take a better look. Seems Haskell doesn't have laws for `MonadError` at all.

posts/2018-04-13-rethinking-monaderror.md
```diff
@@ -0,0 +1,463 @@
...
+
+```scala
+def raiseErrorStops(e: E, a: A): Boolean =
+  F.raiseError[A](e).flatMap(_ => a.pure[F]) === F.raiseError[A](e)
+```
+
+In the same way, we want values of `G` to never stop propagating, so `flatMap` should always work:
+
+```scala
+def flatMapAlwaysWorks(ga: G[A], a: A): Boolean =
```
edmundnoble: I don't understand this; this seems to be the exact opposite of the monad law for `flatMap` and `pure`. Except for the extra `a`, which actually makes this law guaranteed by parametricity as well for any monad other than the `Void[A] = Nothing` monad.

edmundnoble: The last bit there I'm not sure about. This law just confuses me.

posts/2018-04-13-rethinking-monaderror.md
```diff
@@ -0,0 +1,463 @@
...
+tut:
+  scala: 2.12.4
+  binaryScala: "2.12"
+  dependencies:
+    - org.scala-lang:scala-library:2.12.4
+    - org.typelevel::cats-core:1.1.0
+
+---
+
+`MonadError` is a very old type class, hackage shows me it was originally added in 2001, long before I had ever begun doing functional programming.
```
gvolpe: Might be nice to add the hackage link for reference?

gvolpe:
Very well written and easy to follow :clap: 

I think it should be great to have an extension of `MonadError` or something like `UIO` as defined in `Scalaz 8 IO` to be able to rely on the type system, and I think the `Option` example with nested `attempt` calls emphases the problem very clearly.

This is one of the reasons why we haven't enforced the use of `MonadError` at work and we are still relying on types such as `F[Error \/ A]` even though we must pay a performance and messier code penalty of using Monad Transformers such as `EitherT`.

posts/2018-04-13-rethinking-monaderror.md
LukaJCB: Yeah, as I said above, it's not really satisfying right now. I just want a law that prevents an error-ful type in `G`. Without this law I can define instances like `MonadBlunder[Task, Task, Throwable]`, which sort of defeats the point I guess.

posts/2018-04-13-rethinking-monaderror.md
```diff
@@ -0,0 +1,465 @@
...
+The implementation is fairly straightforward as well, we just handle all the errors my lifting them into the left side of an `Either` and map successful values to the right side of `Either`.
+
+Next, let's look at the dual to `attempt`, called `rethrow` in Cats. 
+For `MonadError` it turns an `F[Either[E, A]]` back into an `F`, but we're going to use our unexceptional type again:
+
+```scala
+trait MonadBlunder[F[_], G[_], E] {
+  ...
+
+  def absolve[A](fa: F[A]): G[Either[E, A]] = ???
```
smarter: Wrong signature (the correct one is used below)

LukaJCB: Woops, thank you!

posts/2018-04-13-rethinking-monaderror.md
edmundnoble: `flatMap` has to do nothing anyway, because of the law that states `flatMap(_.pure[F])` does nothing, as well as because `flatMap` can be specified in terms of `map` and `join`.

LukaJCB: Good point, I wonder if Phil thought of that or had something else in mind.

posts/2018-04-13-rethinking-monaderror.md
Blaisorblade: To express what you want, I think you want a typeclass method that enables _injecting_ `G` into `F` (`def inject[A](g: G[A]): F[A]`) — you seem to imply that's possible anyway. Then you can try to state that handling errors in `inject(g)` never has anything to handle — one way to write that might be `handleErrorWith(inject(ga))(f) = ga`. @LukaJCB @edmundnoble WDYT?

LukaJCB: That's excellent, thank you!

PR #210, opened 2018-05-09
Blog post: Tagless Final algebras and Streaming

gvolpe:
Also added details about author since it's my first time writing a blog post here :)


posts/2018-05-09-tagless-final-streaming.md
```diff
@@ -0,0 +1,172 @@
...
+
+// Pretend this is the real implementation... What matters for the example are the types.
+class PostgreSQLItemRepository[F[_]](implicit F: Sync[F]) extends ItemRepository[F] {
+  private val mockItem = Item(ItemName("laptop"), BigDecimal(1500))
+
+  override def findAll: F[List[Item]] = F.pure(List(mockItem))
+  override def find(name: ItemName): F[Option[Item]] = F.pure(Some(mockItem))
+  override def save(item: Item): F[Unit] = F.unit
+  override def remove(name: ItemName): F[Unit] = F.unit
+}
```
oleg-py: This makes a poor example for having `Sync` constraint. Even if it's just a stub, you can add some `println`s to show that there are effects you suspend. Right now it only requires `Applicative` methods

Especially considering a latter point about abstracting over effect types.

posts/2018-05-09-tagless-final-streaming.md
```diff
@@ -0,0 +1,172 @@
...
+import cats.Monad
+
+def fp[F[_]: Monad]: F[String] =
+`  for {
+    a <- bar[F]
+    b <- bar[F]
+  } yield a + b
+```
+
+The above implementation makes use of a `for-comprehention` which is a syntactic sugar for `flatMap` and `map`, so all we need is a `Monad[F]` instance because we also need an `Applicative[F]` instance for `bar`, otherwise we could just use a `FlatMap[F]` instance.
```
oleg-py: This is also a so-so example. `(bar[F], bar[F]).mapN(_ + _)` would only require `Applicative` and because of consistency law would also work for any `Monad` in exactly the same way.

posts/2018-05-09-tagless-final-streaming.md
```diff
@@ -0,0 +1,172 @@
...
+
+There have been a couple of really [nice blog posts](https://typelevel.org/blog/2017/12/27/optimizing-final-tagless.html) about `Tagless Final` and some related topics. However, I have faced some design problems when writing some algebras and haven't seen anybody talking about. So please let me introduce this problem to you.
+
+### Algebra definition
+
+Consider the following algebra:
+
+```scala
+trait ItemRepository[F[_]] {
+  def findAll: F[List[Item]]
```
oleg-py: Just for completeness sake, I'd add the definition of `class Item` above that.

posts/2018-05-09-tagless-final-streaming.md
```diff
@@ -0,0 +1,172 @@
...
+Now it could be very common to have a huge amount of `Item`s in our DB that a `List` will not fit into memory and / or it will be a very expensive operation. So we might want to stream the results of `findAll` instead of have them all in memory on a `List`, making `Doobie` a great candidate for the job. But wait... We have a problem now. Our `ItemRepository` algebra has fixed the definition of `findAll` as `F[List[Item]]` so we won't be able to create an interpreter that returns a streaming result instead.
+
+### Rethinking our algebra
+
+We should think about abstracting over that `List` and two of the most common abstractions that immediately come to mind are `Foldable` and `Traverse`. But although these typeclasses are very useful, they are not enough to represent a stream of values, so we should come up with a better abstraction.
+
+Well, it seems that our options are either adding another higher-kinded parameter `G[_]` to our algebra or just define an abstract member `G[_]`. So let's go with the first one:
+
+```scala
+trait ItemRepository[F[_], G[_]] {
```
mpilquist: Would be interesting to compare this to `trait ItemRepository[F[_], S[_[_], _]] { def findAll: S[F, Item] }`

posts/2018-05-09-tagless-final-streaming.md
```diff
@@ -0,0 +1,172 @@
...
+  override def find(name: ItemName): Id[Option[Item]] = mem.get(name.value)
+  override def save(item: Item): Id[Unit] = mem.update(item.name.value, item)
+  override def remove(name: ItemName): Id[Unit] = {
+    mem.remove(name.value)
+    ()
+  }
+}
+```
+
+That's pretty much it! We managed to abstract over the return type of `findAll` by just adding an extra parameter to our algebra.
```
oleg-py: This is nice and all, but I don't see how the pieces would compose together. Say, I want to write a generic function which takes all items (using `findAll`), applies some discounts (I'm assuming the second parameter of `Item` is a price) and writes them back using `save`. If I specified `Stream` or `List`, it'd be immediately obvious how to do that, but not with that second type parameter which I cannot quite constraint. Such an example would be very interesting.

posts/2018-05-09-tagless-final-streaming.md
gvolpe: At first sight, the problem I see with this signature is how to create an interpreter where `findAll: List[Item]` or `findAll: F[List[Item]]`? Seems to be more specific to the shape of `fs2.Stream`. By having `F[_], G[_]` I can fix `G` to `Stream[F, ?]` and it also works for types of kind `* -> *` such as `List`, `Vector`, etc.

posts/2018-05-09-tagless-final-streaming.md
gvolpe: Fair point, I guess I was a bit lazy with this example, will update to have at least `F.delay(println(s"saving item: $item"))` :smile:

posts/2018-05-09-tagless-final-streaming.md
Wogan: typo: `for-comprehension`

posts/2018-05-09-tagless-final-streaming.md
mpilquist: `Stream[Id, ?]`

mpilquist: `Stream[Id, ?]`

posts/2018-05-09-tagless-final-streaming.md
gvolpe: The idea here was to show that by using a for-comprehention `flatMap` is required, hence one will need either a `FlatMap` or `Monad` constraint. I understand that with `mapN` only an `Applicative` is needed but the idea here is a bit different, not sure if I get your point?

gvolpe: Thanks, I always spell this one wrong :smile:

posts/2018-05-09-tagless-final-streaming.md
```diff
@@ -0,0 +1,172 @@
...
+
+### Source of inspiration
+
+I've come up with most of the ideas presented here during my work on [Fs2 Rabbit](https://gvolpe.github.io/fs2-rabbit/), a stream based client for `Rabbit MQ`, where I make heavy use of this technique as I originally described in [this blog post](https://partialflow.wordpress.com/2018/02/01/a-tale-of-tagless-final-cats-effect-and-streaming-fs2-rabbit-v0-1/).
+
+Another great source of inspiration was [this talk](https://www.youtube.com/watch?v=1h11efA4k8E) given by [Luka Jacobowitz](https://github.com/LukaJCB) at Scale by the Bay.
```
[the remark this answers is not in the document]
gvolpe: http://scale.bythebay.io/

posts/2018-05-09-tagless-final-streaming.md
gvolpe: Fair point. I actually have a bigger example putting all the pieces together and showing that to integrate the streaming `findAll` with `save` you'll need to compile that streaming to an `F` using `compile`, will try to fit it in the post whilst trying to keep it concise.

gvolpe: I just re-read your comment (missed the "generic function" part before). I'll try to add a short section of why I wouldn't recommend this and strictly do it at the interpreter level instead.

posts/2018-05-09-tagless-final-streaming.md
gvolpe: That's an interesting perspective. I'll mention it in this post as another option for the final user, but I'd prefer to avoid diving into more details and leave it for another post :smiley:

mpilquist: :) Sounds good to me. I think what bother me about F and G is that their relationship is completely unconstrained. I want to know that F is my effect type and S is my stream type.

posts/2018-05-09-tagless-final-streaming.md
```diff
@@ -204,6 +204,33 @@ While in this case it was possible to make it generic I don't recommend to do th
...
+
+Where the second type parameter matches the shape of `fs2.Stream`. In this case our streaming repository will remain the same (it should just extend `ItemRepository[F, Stream]` instead of `ItemRepository[F, Stream[F, ?]]`) but our in memory interpreter will now rely on `fs2.Stream` instead of a parametric `G[_]`, for example:
+
+```scala
+object MemRepositoryAlt extends ItemRepository[Id, Stream] {
+
+  override def findAll: Stream[Id, Item] = Stream.emits {
+    sql"select name, price from items"
+      .query[Item]
+      .to[List]
```
mpilquist: Instead of calling .to[List], call .stream — then no need for the #tream.emits call and everything is streamed from DB.

posts/2018-05-09-tagless-final-streaming.md
gvolpe: Sure, in the algebra there's nothing saying that whenever `G[_]` is a `Stream[_[_], _]` the first parameter of that stream should be `F[_]` but I think that could be seen as an advantage as well. The idea of tagless final algebras is to know very little and almost nothing about the final types.

Anyway, I like this design alternative and I'll probably explore it further. Just added a section in this post with it, thanks!

posts/2018-05-09-tagless-final-streaming.md
gvolpe: You're absolutely right :D

gvolpe: I did it in the `StreamingRepository` and didn't realize about this one :p

posts/2018-05-09-tagless-final-streaming.md
```diff
@@ -0,0 +1,285 @@
...
+
+  def process(discount: Double): F[Unit] = {
+    val rs: G[Item] = repo.findAll.map(item => item.copy(price = item.price * (1 - discount)))
+    // And now what? repo.save returns F[Unit]
+  }
+
+}
+```
+
+First of all, there's a fundamental problem. We are assuming that all the items are a finite number that fit into memory.
```
fthomas: I don't follow where we have made this assumption. With concrete types we certainly can find all `Item`s, modify them, and write them back (e.g. via `evalMap`) in one `Stream` without collecting all `Itme`s into memory.

gvolpe: That's right. I was thinking about List and Traverse specifically when I wrote that sentence but now I see it's in the wrong position and does not make sense, will remove it.

_data/authors.yml
```diff
@@ -309,3 +309,10 @@ annettebieniusa:
   full_name: "Annette Bieniusa"
   portrait: "/img/media/speakers/annettebieniusa.jpg"
   bio: "Annette is a lecturer and senior researcher at the Technische Universität Kaiserslautern. Her research interests include semantics of concurrent and distributed programming, with a focus on replication, synchronization, and how they are reflected on programming language level. Annette was involved in several national and international research projects, most recently the in the EU-Projects “SyncFree: Large-scale Computation without Synchronization” and “Lightkone: Lightweight computation for networks at the edge“."
+gvolpe:
+  full_name: "Gabriel Volpe"
+  portrait: "/img/media/speakers/gvolpe.jpg"
+  github: "gvolpe"
+  twitter: "volpegabriel87"
+  bio: "FP developer Scala | Haskell"
```
larsrh: Bio is not necessary, it will not be rendered below the blog post.

gvolpe: Okay, will remove. Copy / pasted from others :smile:

posts/2018-05-09-tagless-final-streaming.md
```diff
@@ -0,0 +1,286 @@
...
+                                                   fk: FunctionK[G, λ[x => F[H[x]]]]) {
+
+  def process(discount: Double): F[Unit] = {
+    val items: G[Item] = repo.findAll.map(item => item.copy(price = item.price * (1 - discount)))
+    val fa: F[H[Item]] = fk.apply(items)
+    fa.flatMap(t => t.traverse(item => repo.save(item)).void)
+  }
+
+}
+
```
fthomas: I'm wondering if something simpler like
```scala
class DiscountProcessor[F[_], G[_]: Functor](repo: ItemRepository[F, G], join: G[F[Unit]] => F[Unit]) {

  def process(discount: Double): F[Unit] = {
    val items: G[Item] = repo.findAll.map(item => item.copy(price = item.price * (1 - discount)))
    val saved: G[F[Unit]] = items.map(repo.save)
    join(saved)
  }
}
```
would work for `StreamingDiscountInterpreter` and `ListDiscountInterpreter` without requiring to load all items in memory in `StreamingDiscountInterpreter`.

gvolpe: I'm curious, how would you define `join`?

fthomas: For `StreamingDiscountInterpreter` its signature would be `Stream[IO, IO[Unit]] => IO[Unit]`, right? So `_.evalMap(identity).compile` should fit the types.

gvolpe: You're completely right and have destroyed that section :smile: 

I need to put a bit more of thought on it to see how other situations could be affected, but from a first sight it looks very correct, will be updating it soon, thanks!

gvolpe: Updated it @fthomas , thanks again :+1:

gvolpe:
Came here to share my thoughts and ended up learning quite a lot, thanks to all the reviewers that have taken the time to go through the article, awesome community :heart:

larsrh:
I've resolved the merge conflict in the authors file. Any other pending changes, or is this good to merge?

gvolpe:
Thanks @larsrh ! No more changes from my side. @fthomas are you okay with the latest change suggested by you?

fthomas:
@gvolpe Yes, LGTM.

larsrh:
:shipit:

larsrh:
@gvolpe Please remember to subscribe to the comment thread: https://typelevel.org/blog/2018/05/09/tagless-final-streaming.html

PR #211, opened 2018-05-10
Add Product with Serializable blog post

ceedubs:
This is a document that I've had sitting on my laptop for a while. I
figured that I might as well put it in a public place, and the typelevel
blog is the first place that came to mind.

I now see that Noel Welsh has published something similar [here](https://underscore.io/blog/posts/2015/06/04/more-on-sealed.html).
This goes into a bit more detail with type LUBs, which may be helpful to
someone?

### DIFF
--- a/posts/2018-05-09-product-with-serializable.md
+++ b/posts/2018-05-09-product-with-serializable.md
@@ -15,7 +15,7 @@ tut:
 
 ---
 
-A somewhat common scala idiom is to make an `abstract` type extend `Product with Serializable`. There isn't an obvious reason to do this, and people have asked me a number of times why I've done this. While I don't think that `Product` or `Serializable` are particularly good abstractions, there's a reason that I extend them.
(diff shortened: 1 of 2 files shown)


posts/2018-05-09-product-with-serializable.md
```diff
@@ -0,0 +1,65 @@
...
+
+tut:
+  scala: 2.12.4
+  binaryScala: "2.12"
+  dependencies:
+    - org.scala-lang:scala-library:2.12.4
+
+---
+
+A somewhat common scala idiom is to make an `abstract` type extend `Product with Serializable`. There isn't an obvious reason to do this, and people have asked me a number of times why I've done this. While I don't think that `Product` or `Serializable` are particularly good abstractions, there's a reason that I extend them.
```
larsrh: scala → Scala

ceedubs:
Thanks for the quick review @larsrh! I've pushed an update.

larsrh:
Thanks for the post!

PR #215, opened 2018-06-06
Blog post: Shared State in Functional Programming

gvolpe:
First shared in this [gist](https://gist.github.com/gvolpe/d4facceedf12ab7870934cbf9d6174e9) but here I added a few other things after getting some feedback from Reddit :)

Also thanks to @tpolecat for pushing me to write it as a `Typelevel` blog post.

### DIFF
--- a/posts/2018-06-07-shared-state-in-fp.md
+++ b/posts/2018-06-07-shared-state-in-fp.md
@@ -16,49 +16,9 @@ tut:
 
 ---
 
-Newcomers to Functional Programming are often very confused about the proper way to share state without breaking purity and end up having a mix of pure and impure code that defeats the purpose of having pure FP code in the first place.
+Newcomers to functional programming (FP) are often very confused about the proper way to share state without breaking purity and end up having a mix of pure and impure code that [defeats the purpose](https://queue.acm.org/detail.cfm?id=2611829) of having pure FP code in the first place.
 
 This is the reason that has motivated me to write a beginner friendly guide :)
-
-## Referential Transparency
-
-I think this is the key to understanding the benefits of Functional Programming. Without it, it'll be hard to understand other deeper concepts, such as `Shared State`. So let's start this guide by refreshing our knowledge about it:
-
-##### What is Referential Transparency?
-
-Being able to substitute an expression by its bound value without changing the meaning of the program.
-
-##### What does this mean?
-
-An example is worth thousand words. Consider this:
-
-```scala
-val expr = 123
-(expr, expr)
-```
-
-If we now substitute the expression by its bound value, does the meaning of the program change?
-
-```
-(123, 123)
-```
-
-No! The program remains the same. We call this property *Referential Transparency*. How about this?
-
-```scala
-val expr = println("hey")
-(expr, expr)
-```
-
-Let's perform susbtitution once again:
-
-```scala
-(println("hey"), println("hey"))
-```
-
-M'kay... The meaning of the program has now changed. The first program prints "hey" once whereas the second program prints "hey" twice. This is not `Referential Transparent`.
-
-So having this knowledge, let's now move onto the next section!
 
 ## Use Case
 
@@ -288,7 +248,7 @@ object sharedstate extends IOApp {
 
 As I mentioned in one of the sections above, the creation of `Ref[F, A]` is side-effectful. So what does this mean? Does it write to disk? Does it perform HTTP Requests? Not exactly.
 
-It all comes down to wanting to keep the property of *Referential Transparency* while sharing and mutating state. So let's again put up an example to follow up along with some explanation:
+It all comes down to wanting to keep the property of *referential transparency* while sharing and mutating state. So let's again put up an example to follow up along with some explanation:
 
 ```scala
 var a = 0
@@ -332,7 +292,7 @@ We have now regained purity. So whenever you create an `IORef` you'll get an `IO
 
 ## Applying the technique in other libraries
 
-Although in the example above we only see how it's done with the `Cats Effect` library, this principle expands to other FP libraries as well.
+Although in the example above we only see how it's done with the `cats-effect` library, this principle expands to other FP libraries as well.
 
 For example, when writing an `http4s` application you might need to create an `HttpClient` that needs to be used by more than one of your services. So again, create it at startup and `flatMap` it once:
 
@@ -369,4 +329,4 @@ To conclude this post I would like to give a big shout out to [@SystemFW](https:
 
 > In the specific case of state sharing, this gives rise to a really nice property: since the only way to share is passing things as an argument, *the regions of sharing are exactly the same of your call graph*, so you transform an important aspect of the behaviour ("who shares this state?") into a straightforward syntactical property ("what methods take this argument"?). This makes shared state in pure FP a lot easier to reason about than its side-effectful counterpart imho.
 
-In simple terms, remind yourself about this: **"FlatMap once and pass the reference as an argument!"**
+In simple terms, remind yourself about this: **"flatMap once and pass the reference as an argument!"**
--- a/posts/2018-06-07-shared-state-in-fp.md
+++ b/posts/2018-06-07-shared-state-in-fp.md
@@ -1,0 +1,372 @@
+---
+layout: post
+title: Shared State in Functional Programming
+
+meta:
+  nav: blog
+  author: gvolpe
+  pygments: true
+
+tut:
+  scala: 2.12.4
+  binaryScala: "2.12"
+  dependencies:
+    - org.scala-lang:scala-library:2.12.4
+    - org.typelevel::cats-core:1.1.0
+
+---
+
+Newcomers to Functional Programming are often very confused about the proper way to share state without breaking purity and end up having a mix of pure and impure code that defeats the purpose of having pure FP code in the first place.
+
+This is the reason that has motivated me to write a beginner friendly guide :)
+
+## Referential Transparency
+
+I think this is the key to understanding the benefits of Functional Programming. Without it, it'll be hard to understand other deeper concepts, such as `Shared State`. So let's start this guide by refreshing our knowledge about it:
+
+##### What is Referential Transparency?
+
+Being able to substitute an expression by its bound value without changing the meaning of the program.
+
+##### What does this mean?
+
+An example is worth thousand words. Consider this:
+
+```scala
+val expr = 123
+(expr, expr)
+```
+
+If we now substitute the expression by its bound value, does the meaning of the program change?
+
+```
+(123, 123)
+```
+
+No! The program remains the same. We call this property *Referential Transparency*. How about this?
+
+```scala
+val expr = println("hey")
+(expr, expr)
+```
+
+Let's perform susbtitution once again:
+
+```scala
+(println("hey"), println("hey"))
+```
+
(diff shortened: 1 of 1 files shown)


larsrh:
@adelbertc Mind giving this  a review?

adelbertc:
@gvolpe @larsrh I will be sure to do it tomorrow!

gvolpe:
Great, thanks @adelbertc !

posts/2018-06-07-shared-state-in-fp.md
```diff
@@ -0,0 +1,372 @@
...
+tut:
+  scala: 2.12.4
+  binaryScala: "2.12"
+  dependencies:
+    - org.scala-lang:scala-library:2.12.4
+    - org.typelevel::cats-core:1.1.0
+
+---
+
+Newcomers to Functional Programming are often very confused about the proper way to share state without breaking purity and end up having a mix of pure and impure code that defeats the purpose of having pure FP code in the first place.
```
adelbertc: No need to capitalize "functional programming" and consider parenthesizing (FP) next to it so your future use of FP is clear

posts/2018-06-07-shared-state-in-fp.md
```diff
@@ -0,0 +1,372 @@
...
+
+This is the reason that has motivated me to write a beginner friendly guide :)
+
+## Referential Transparency
+
+I think this is the key to understanding the benefits of Functional Programming. Without it, it'll be hard to understand other deeper concepts, such as `Shared State`. So let's start this guide by refreshing our knowledge about it:
+
+##### What is Referential Transparency?
+
+Being able to substitute an expression by its bound value without changing the meaning of the program.
```
adelbertc: The use of "bound value" might suggest this only applies to variable assignment/capture, in general RT says a given expression *e* is referentially transparent if you can evaluate it and substitute its value in **any** larger expression it appears in without changing the meaning of the larger expression.

posts/2018-06-07-shared-state-in-fp.md
```diff
@@ -0,0 +1,372 @@
...
+(expr, expr)
+```
+
+If we now substitute the expression by its bound value, does the meaning of the program change?
+
+```
+(123, 123)
+```
+
+No! The program remains the same. We call this property *Referential Transparency*. How about this?
```
adelbertc: No need to capitalize referential transparency.

posts/2018-06-07-shared-state-in-fp.md
```diff
@@ -0,0 +1,372 @@
...
+(expr, expr)
+```
+
+Let's perform susbtitution once again:
+
+```scala
+(println("hey"), println("hey"))
+```
+
+M'kay... The meaning of the program has now changed. The first program prints "hey" once whereas the second program prints "hey" twice. This is not `Referential Transparent`.
```
adelbertc: No need to capitalize

posts/2018-06-07-shared-state-in-fp.md
adelbertc: This might be a good article to cite https://queue.acm.org/detail.cfm?id=2611829

posts/2018-06-07-shared-state-in-fp.md
```diff
@@ -0,0 +1,372 @@
...
+
+### Getting started
+
+So this is how we might decide to start writing our code:
+
+```scala
+import cats.effect._
+import cats.effect.concurrent.Ref
+import cats.instances.list._
+import cats.syntax.all._
```
adelbertc: You can replace these last 2 imports with just `import cats.implicits._`

posts/2018-06-07-shared-state-in-fp.md
```diff
@@ -0,0 +1,372 @@
...
+List(#2, #1, #3)
+```
+
+We'll use the concurrency primitive `Ref[IO, List[String]]` to represent our internal state because it's a great fit.
+
+### Getting started
+
+So this is how we might decide to start writing our code:
+
+```scala
```
adelbertc: Consider using `tut` here so it actually goes through the type checker in CI http://tpolecat.github.io/tut/

Going to stop writing this for the rest of the review, but there's a couple code blocks below that can also benefit

posts/2018-06-07-shared-state-in-fp.md
```diff
@@ -0,0 +1,372 @@
...
+
+If we now substitute the expression by its bound value, does the meaning of the program change?
+
+```
+(123, 123)
+```
+
+No! The program remains the same. We call this property *Referential Transparency*. How about this?
+
+```scala
```
adelbertc: Consider using `tut`

posts/2018-06-07-shared-state-in-fp.md
```diff
@@ -0,0 +1,372 @@
...
+
+##### What is Referential Transparency?
+
+Being able to substitute an expression by its bound value without changing the meaning of the program.
+
+##### What does this mean?
+
+An example is worth thousand words. Consider this:
+
+```scala
```
adelbertc: Consider using `tut`

posts/2018-06-07-shared-state-in-fp.md
```diff
@@ -0,0 +1,372 @@
...
+val expr = 123
+(expr, expr)
+```
+
+If we now substitute the expression by its bound value, does the meaning of the program change?
+
+```
```
adelbertc: Consider using `tut`

posts/2018-06-07-shared-state-in-fp.md
```diff
@@ -0,0 +1,372 @@
...
+No! The program remains the same. We call this property *Referential Transparency*. How about this?
+
+```scala
+val expr = println("hey")
+(expr, expr)
+```
+
+Let's perform susbtitution once again:
+
+```scala
```
adelbertc: Consider using `tut`

posts/2018-06-07-shared-state-in-fp.md
```diff
@@ -0,0 +1,372 @@
...
+    masterProcess.as(ExitCode.Success)
+
+}
+```
+
+## Mutable reference
+
+As I mentioned in one of the sections above, the creation of `Ref[F, A]` is side-effectful. So what does this mean? Does it write to disk? Does it perform HTTP Requests? Not exactly.
+
+It all comes down to wanting to keep the property of *Referential Transparency* while sharing and mutating state. So let's again put up an example to follow up along with some explanation:
```
adelbertc: no need to capitalize

posts/2018-06-07-shared-state-in-fp.md
```diff
@@ -0,0 +1,372 @@
...
+}
+```
+
+We have now regained purity. So whenever you create an `IORef` you'll get an `IO[IORef]` instead of a mutable reference to `IORef`. This means that when you invoke `flatMap` on it twice you'll get two different mutable states, and this is the power of `Referential Transparency`. It gives you way more control than having a `val ref` hanging around in your code and gives you ***local reasoning***.
+
+*All these examples are written in terms of `IO` for the sake of simplicity but in practice they are polymorphic on the effect type.*
+
+## Applying the technique in other libraries
+
+Although in the example above we only see how it's done with the `Cats Effect` library, this principle expands to other FP libraries as well.
```
adelbertc: `cats-effect`

posts/2018-06-07-shared-state-in-fp.md
```diff
@@ -0,0 +1,372 @@
...
+
+> At the end of the day all the benefits from *referential transparency* boil down to being able to understand and build code compositionally. That is, understanding code by understanding individual parts and putting them back together, and building code by building individual parts and combining them together. This is only possible if *local reasoning* is guaranteed, because otherwise there will be weird interactions when you put things back together, and referential transparency is *defined* as something that guarantees local reasoning.
+
+> In the specific case of state sharing, this gives rise to a really nice property: since the only way to share is passing things as an argument, *the regions of sharing are exactly the same of your call graph*, so you transform an important aspect of the behaviour ("who shares this state?") into a straightforward syntactical property ("what methods take this argument"?). This makes shared state in pure FP a lot easier to reason about than its side-effectful counterpart imho.
+
+In simple terms, remind yourself about this: **"FlatMap once and pass the reference as an argument!"**
```
adelbertc: `flatMap`

posts/2018-06-07-shared-state-in-fp.md
larsrh: There's even a `tut` header in the file :thinking:

gvolpe:
Thanks for the feedback @adelbertc ! Maybe I should leave the "Referential Transparency" section out and aim for the second type of audience as the [original post](https://gist.github.com/gvolpe/d4facceedf12ab7870934cbf9d6174e9). I'll be addressing the comments soon.

gvolpe:
Just give a bit more of context, this post is aimed for those that are currently making use of `IO` or `Task`, understand basic operations such as `flatMap` and `map` or even more but still don't feel super comfortable doing pure FP and end up mixing it with mutable state and a few `unsafeRunSync()`.

posts/2018-06-07-shared-state-in-fp.md
gvolpe: Yes, but tbh I prefer to be explicit about the imports

SethTisue:
> Just to give a bit more of context, this post is aimed for those that are currently making use of IO or Task, understand basic operations such as flatMap and map or even more but still don't feel super comfortable doing pure FP and end up mixing it with mutable state and a few unsafeRunSync().

succinctly put. say so at the top of the blog post itself...?

posts/2018-06-07-shared-state-in-fp.md
gvolpe: I'm :-1: on using `tut` for a blog post. I think it's great for docs to keep it updated but not really for a blog post. 

The reason being that to preview this blog post I must compile all the existent blog posts and we might have divergence in the future when updating dependencies and an old post that was completely valid back in time is no longer valid, even if we force it to compile, because also the sentences will need to be updated.

gvolpe:
@SethTisue I think that would probably be too much but I did add links to `IO` and `Ref`, and subtly suggested that some knowledge of Cats Effect is required in order to understand the examples in the Getting Started section.

larsrh:
@gvolpe 

> The reason being that to preview this blog post I must compile all the existent blog posts

You'll have to do that anyway, even if this post doesn't use it.

> we might have divergence in the future when updating dependencies and an old post that was completely valid back in time is no longer valid, even if we force it to compile, because also the sentences will need to be updated.

No, that's not a risk. The `tut` header specifies a fixed set of dependencies per file.

In any case, there's no need for your post to use it, but in that case, please remove the header.

gvolpe:
Thanks for correcting me @larsrh , my bad, didn't pay attention to the header. I'll try to get it compiling with `tut` as suggested :+1:

gvolpe:
I actually caught some bad code using `tut`, good advice :smile: In some parts I didn't use it either because of some limitations of `tut` (eg `private class`) or for convenience (eg. the `http4s` example at the end).

larsrh:
@adelbertc Did @gvolpe address all your comments? If so, please update your review to green.

larsrh:
@gvolpe Thanks for your contribution!

PR #216, opened 2018-06-14
new post: typedapi

larsrh:
Post as submitted @pheymann.

I haven't changed much (in particular, I haven't proofread it). Reviews most welcome!

### DIFF
--- a/posts/2018-06-15-typedapi.md
+++ b/posts/2018-06-15-typedapi.md
@@ -1,0 +1,329 @@
+---
+layout: post
+title: Typedapi or how to derive your clients and servers from types
+
+meta:
+  nav: blog
+  author: pheymann
+  pygments: true
+---
+
+In this blog post, I will show you how to leverage Scala's type system to derive an HTTP client function from a single type. This will also be the story of how I started to work on [Typedapi](https://github.com/pheymann/typedapi) which is basically the attempt to bring Haskell's [Servant](https://github.com/haskell-servant/servant) to Scala.
+
+## Servant in a nutshell and how it began
+For everyone not knowing Servant, it is a library which lets you define your web apis as types and derives the client and server functions from it. When I saw it for the first time while working on a ped project I immediately loved the idea. Creating web server and clients this way reduces your code to a mere type, you get extra type safety and you can use the api types as contracts between your server and its clients. 
+
+I couldn't find any viable alternative in Scala at the time and decided to build it on my own. But I just wanted to start with a single feature to not overwhelm myself and abandon the project after a short time. Therefore, I set out to make Scala able to derive a client function from a single api type, as we will do in this post.
+
+## Derive a client function from a type. How hard can it be?
+Let's start with an example we will use later on to ease understanding. Consider the following api:
+
+```
+GET /users/:name?minAge=:age -> List[User]
+```
+
+It only consists of a single endpoint which returns a list of `Users`:
+
+```scala
+final case class User(name: String, age: Int)
+```
+
+with a given `name: String`. Furthermore, you filter the resulting users by their `age: Int`. Our big goal is to end up with a function which is derived from a type-level representation of our endpoint:
+
+```scala
+(name: String, minAge: Int) => F[List[User]]
+```
+
+### Represent the api as a type
+Question: how do you represent the above api as a type in Scala? I think the best way is to break it apart and try to find type-level representations for each element. After that, we "just" merge them together.
+
+When we take a closer look at our endpoint we see that it consists of:
+  * path elements identifying an endpoint: `/users`
+  * segments within the path: `:name`
+  * queries: `minAge=[age]`
+  * and a method `GET` to identify which kind of operation we want to do
+
+Or in other words, just a plain HTTP definition of a web endpoint. Now that we know what we are working with let's try and find a type-level representation.
+
+We start with our path element. If we would work with Dotty or Typelevel-Scala we could use a feature called literal-type:
+
+```scala
+type Path = "users"
+```
+
+But as we want to stay in Vanilla Scala this isn't an option.
+We have to use the one tool probably every developer has to use when it comes to working on the type-level called [shapeless](https://github.com/milessabine/shapeless).
+It has this nifty class [Witness](https://github.com/milessabin/shapeless/blob/shapeless-2.3.3/core/src/main/scala/shapeless/singletons.scala#L32) which generates a witness type from a literal like `Strings` or `Symbols`.
+
+```scala
+import shapeless.Witness
+
+val usersW = Witness("users")
+```
+
+But this isn't a pure type declaration, you will say. And you are right, but right now there is no other way in Scala. We have to go the ordinary value road first to create our types.
+
+Now that we know how to get a type representation from a `String` which describes our path we should clearly mark it as a path element:
+
+```scala
+sealed trait Path[P]
+
+type users = Path[usersW.T]
+```
+
+That's it. That is the basic concept of how we can describe our apis as types. We just reuse this concept now for the remaining elements like the segment.
+
+
+```scala
+val nameW = Witness('name)
+
+sealed trait Segment[K, V]
+
+type name = Segment[nameW.T, String]
+```
+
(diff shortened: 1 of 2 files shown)


posts/2018-06-15-typedapi.md
```diff
@@ -0,0 +1,329 @@
...
+```scala
+(name: String, minAge: Int) => F[List[User]]
+```
+
+### Represent the api as a type
+Question: how do you represent the above api as a type in Scala? I think the best way is to break it apart and try to find type-level representations for each element. After that, we "just" merge them together.
+
+When we take a closer look at our endpoint we see that it consists of:
+  * path elements identifying an endpoint: `/users`
+  * segments within the path: `:name`
```
vil1: For segments and queries, it might be worth emphasizing that we need them to represent a parameter which has a name *and* a type

posts/2018-06-15-typedapi.md
```diff
@@ -0,0 +1,329 @@
...
+We have to use the one tool probably every developer has to use when it comes to working on the type-level called [shapeless](https://github.com/milessabine/shapeless).
+It has this nifty class [Witness](https://github.com/milessabin/shapeless/blob/shapeless-2.3.3/core/src/main/scala/shapeless/singletons.scala#L32) which generates a witness type from a literal like `Strings` or `Symbols`.
+
+```scala
+import shapeless.Witness
+
+val usersW = Witness("users")
+```
+
+But this isn't a pure type declaration, you will say. And you are right, but right now there is no other way in Scala. We have to go the ordinary value road first to create our types.
```
vil1: If I didn't knew about `Witness` and singleton types, this wouldn't be very clear to me. Maybe it would be a little clearer it you flipped the reasonning like so: 
* the challenge here is to "transform a value into a type" 
* of course we can do that only if the value is already known at compile time (eg a literal)
* Dotty can do exactly that, but in Scala we need to use `Witness`
* a `Witness` *value* "contains" an abstract type `T` that is exactly what we want here

posts/2018-06-15-typedapi.md
```diff
@@ -0,0 +1,329 @@
...
+
+```scala
+val minAgeW = Witness('minAge)
+
+sealed trait Query[K, V]
+
+type minAge = Query[minAgeW.T, Int]
+
+sealed trait Method
+sealed trait Get[A] extends Method
```
vil1: It's worth mentioning that the `A` type parameter will represent the type of the resource returned by the endpoint.

posts/2018-06-15-typedapi.md
```diff
@@ -0,0 +1,329 @@
...
+final case class PathList[P <: HList]() {
+  
+  def /[S](path: Witness.Lt[S]): PathList[S :: P] = PathList()
+  ...
+}
+
+val Root = PathList[HNil]()
+```
+  * Same is true for `QueryList`.
+  * The last step is to merge all these `HLists` types into a single one. Shapeless comes again with a handy type class called `Prepend` which provides us with the necessary functionality. Two `HList` types go in, a single type comes out. And again, we use a type carrier here to store the api type.
```
vil1: I prefer to think about typeclasses (when used for typelevel programming like here) as constraints rather than functions. So I'd say that `Prepend.Aux[Q, P, Api]` makes sure (or constrains) that `Api` is indeed the concatenation of `P` and `Q`. But that's a personal POV.

posts/2018-06-15-typedapi.md
```diff
@@ -0,0 +1,329 @@
...
+
+That's it. We say what goes in and what comes out. You need some examples to get a better idea? Here you go:
+
+```scala
+implicit def pathTransformer[P, El <: HList, KIn <: HList, VIn <: HList, M, Out] = 
+  FoldLeftFunction[Path[P], (El, KIn, VIn, M, Out)] { type Out = (P :: El, KIn, VIn, Out) }
+
+implicit def segmentTransformer[K <: Symbol, V, El <: HList, KIn <: HList, VIn <: HList, M, Out] = 
+  FoldLeftFunction[Segment[K, V], (El, KIn, VIn, M, Out)] { type Out = (SegmentInput :: El, K :: KIn, V :: VIn, Out) }
+
```
vil1: A short explanation in English might help here, for people who're not used to the Scala typelevel programming idioms

posts/2018-06-15-typedapi.md
```diff
@@ -0,0 +1,329 @@
...
+And another hint, you can start with `Nothing` as the initial type for your aggregate.
+
+#### Collect all the request data
+We folded our api type into the new representation. Now we use that to build a function which collects all the data necessary to make a request.
+
+```scala
+type Uri     = List[String]
+type Queries = Map[String, List[String]]
+
```
vil1: again, a short description in English of what these types represent might be helpful, especially stressing the fact that these live in the value level (that isn't a correct way to put it, but hopefully you see what I mean).

posts/2018-06-15-typedapi.md
```diff
@@ -0,0 +1,329 @@
...
+When we apply both solutions we end up with:
+
+```scala
+def derive[H <: HList, Fold, El <: HList, KIn <: HList, VIn <: HList, M, Out]
+  (apiList: ApiTypeCarrier[H])
+  (implicit fold: Lazy[TypeLevelFoldLeft.Aux[H, Nothing, (El, KIn, VIn, M, Out)]],
+            builder: RequestDataBuilder[El, KIn, VIn],
+            vinToFn: FnFromProduct[VIn => ExecutableDerivation[El, KIn, VIn, M, Out]]): vinToFn.Out = 
+  vinToFn.apply(input => new ExecutableDerivation[El, KIn, VIn, M, Out](builder, input))
+```
```
vil1: If I was the author of this post, I would probably slip a joke there about how much bigger the signature of this function is compared to its body, which is exactly what we're looking for when doing typelevel programming :)

posts/2018-06-15-typedapi.md
```diff
@@ -0,0 +1,329 @@
...
+      next(inputs.tail, uri, Map(wit.value.name -> List(inputs.head.toString())) ++ queries)
+  }
+```
+
+The other cases are looking quite similar and it is up to the interested reader to find the implementations.
+
+What we end up with is a nested function call structure which will take an `HList` and returns the `uri` and `queries`.
+
+```scala
+"joe" :: 42 :: HNil => (List("users", "joe"), Map("minAge" -> List("42")))
```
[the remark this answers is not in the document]
pheymann: we only expect `String` and `Int` as inputs. `"users"` is part of the path and known at compile time. For a better understanding have a look at the generated function signature:
 
```Scala
// expected (final)
String => Int => (List[String], Map[String -> List[String]])

// current shape
type Input <: HList

Input => (List[String], Map[String -> List[String]])
```

Maybe I should add a statement to clarify that.

posts/2018-06-15-typedapi.md
pheymann: True, when you use a concept on a (more or less) daily basis you easily forget that it might be not that obvious for someone else why it should/can be used here. 

I will rewrite it to make it more clear.

posts/2018-06-15-typedapi.md
pheymann: My 2 Cents: it depends on what I try to achieve. If I want to proof something like `implicit ev: Monad[F]` it constraints the (a at some point known) type. But when computing something on the type level (like the concatenation) I would tend to describe it as function (computation), because here `Api` is solely determined by this computation.

But yes, in the end you could also describe it as a constraint put on `Api` :).

posts/2018-06-15-typedapi.md
pheymann: tuff task ... me finding a good joke 😄 . I will try to come up with one.

pheymann:
thanks @larsrh for preparing the PR @vil1 for proof reading it. I will overhaul my post once I am back at home and have a proper notebook and internet connection at my disposal. Expect incoming changes Sunday night or Monday morning.

pheymann:
@larsrh @vil1 I started a new branch to apply my changes:

https://github.com/pheymann/typelevel.github.com/blob/post/typedapi/posts/2018-06-15-typedapi.md

I will open a new PR (which will supersede this one) once I have finalized my post (hopefully today).

posts/2018-06-15-typedapi.md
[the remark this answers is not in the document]
vil1: That's probably worth adding these signatures somewhere indeed

PR #217, opened 2018-06-18
new post: typedapi / supersedes #216

pheymann:
This post PR supersedes #216 and takes @vil1 comments into account.

### DIFF
--- a/posts/2018-06-15-typedapi.md
+++ b/posts/2018-06-15-typedapi.md
@@ -51,7 +51,7 @@ But how do you transform a value-level information as a type? First of all, the 
 type Path = "users"
 ```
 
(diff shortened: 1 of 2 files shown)


larsrh:
@vil1 WDYT?

vil1:
I was meaning to take a look today, but had to brawl with a flaky integration test instead. I'll try to review this before midnight (CET).

posts/2018-06-15-typedapi.md
```diff
@@ -0,0 +1,342 @@
...
+  FoldLeftFunction[Segment[K, V], (El, KIn, VIn, M, Out)] { type Out = (SegmentInput :: El, K :: KIn, V :: VIn, Out) }
+```
+
+Here, we get some `Segment` with a name `K` and a type `V` and an intermediate aggregation state we will update by adding a placeholder to `El`, the name to `KIn` and the value type to `VIn`.
+
+Now that we can aggregate types we need a vehicle to traverse our `HList` type and transform it on the fly by using our `FoldLeftFunction` instances. I think yet another type class can help us here.
+
+```scala
+trait TypeLevelFoldLeft[H <: HList, Agg] { type Out }
+
```
vil1: wrapping the following implicit methods in an `object TypeLevelFoldLeft` would enable the reader to copy and use the whole snippet.

pheymann:
@vil1 thanks again for proof-reading my post. Regarding your comment about the conference: this is based on the Typelevel Summit Talk I gave in Berlin in May 😄.

vil1:
Cool! Have you considered proposing this talk to other conferences? ScalaIO might be interested in such presentation for example 😉

PR #218, opened 2018-06-24
Update Alleycats link

taisukeoe:
Since Alleycats was moved into Cats repository.


larsrh:
Thanks, well spotted!

PR #219, opened 2018-06-25
Remove historical events from the home page

propensive:
These events are over, so they should no longer be featured on the home page.


larsrh:
Thanks!

PR #226, opened 2018-07-27
Share state in FP: fix typo

guizmaii:


larsrh:
Thanks!

posts/2018-06-07-shared-state-in-fp.md
```diff
@@ -267,7 +267,7 @@ class IORef {
 }
 ```
 
-This is way better since now the mutation is encapsulated within `IORef` but we are now pushing some resposibility to whoever creates an `IORef`. Consider this:
+This is way better since now the mutation is encapsulated within `IORef` but we are now pushing some responsibility to whoever creates an `IORef`. Consider this:
```
guizmaii: I'm not a native english speaker but for me, `responsibility` should be pluralize here, no ? 

> **some** responsibility

The "some" here indicates that there're more than one responsibility, no ?

larsrh: There's the phrase "to take some responsibility", so plural is not necessary here.

guizmaii: Ok thanks Lars

PR #227, opened 2018-08-07
add blog post: refactoring with monads

mtomko:
I wrote up this short blog post to explain to my coworkers some refactoring I'd done. I mentioned it in passing to @larsrh, who glanced it over and suggested it might be worth reviewing for a wider audience. The whole thing is rather trivial and intended more for beginners than anything.

If Typelevel is interested, I'm happy to contribute it. If not, my feelings won't be hurt, I'll share it with my coworkers as I had originally intended.

If we decide to publish it here, there're a few formatting issues that I'd appreciate a bit of help with. The biggest one is a markdown table that seems to render badly on the Typelevel site but looks fine in plain markdown. I think the default github CSS is fine, but the Typelevel site is perhaps missing some configuration. I was a little nervous to tweak site-wide CSS for fear that I'd ruin something more fundamental.

Similarly, some of the paragraph breaks look awfully long to me, but the markdown file leaves only a single line in between.

### DIFF
--- a/posts/2018-08-07-refactoring-monads.md
+++ b/posts/2018-08-07-refactoring-monads.md
@@ -49,7 +49,6 @@ don't always get guidance on how to achieve that. In the presence of
 error handling and nested data structures, the problem gets even
 harder.
 
-### Breaking up code into smaller pieces
 The goal of this blog post is to describe a concrete strategy for
 structuring code so that the overall flow of control is clear to the
 reader, even months later; and so the smaller pieces are both
@@ -64,12 +63,19 @@ wrap smaller pieces and ultimately, compose them into an
 understandable sequence of computations.
 
 ### Example domain: reading a catalog
+
 Let's not worry about `MonadError` yet, but instead look at some
-example code. Suppose we need to read an object from a relational
-database. Unfortunately, rows in the table may represent objects of a
-variety of types so we have to read the row and build up the object
-graph. This is the boundary between the weakly typed wilderness and
-the strongly typed world within our program.
+example code. Consider a situation where you need to translate data
+from one domain model to another one with different restrictions, and
+controlled vocabularies. This can happen in a number of places in a
+program, for instance reading a database or an HTTP request to
+construct a domain object.
+
+Suppose we need to read an object from a relational database.
+Unfortunately, rows in the table may represent objects of a variety of
+types so we have to read the row and build up the object graph
+accordingly. This is the boundary between the weakly typed wilderness
+and the strongly typed world within our program.
 
 Say our database table represents a library catalog, which might have
 print books and ebooks. We'd like to look up a book by ID and get back
@@ -312,8 +318,6 @@ Here we go:
 ```tut:silent
 import cats.MonadError
 import cats.implicits._
-
-import scala.util.{Failure, Success, Try}
 
 def tryParse[F[_], A](s: String, parse: String => Option[A])(
     implicit me: MonadError[F, Throwable]): F[A] = {
--- a/posts/2018-08-07-refactoring-monads.md
+++ b/posts/2018-08-07-refactoring-monads.md
@@ -260,7 +260,9 @@ def findBookById(id: Int): Try[Book] =
   for {
     row <- DB.unsafeQueryUnique(sql"""select * from catalog where id = $id""")
     format <- Format.fromString(row[String]("format"))
-    (id, title, author) = (row[Int]("id"), row[String]("title"), row[String]("author"))
+    id = row[Int]("id")
+    title = row[String]("title")
+    author = row[String]("author")
     book <- format match {
       case Print =>
         Success(PrintBook(id, title, author))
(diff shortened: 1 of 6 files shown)


larsrh:
> The biggest one is a markdown table that seems to render badly on the Typelevel site but looks fine in plain markdown.

I'll take a look; I don't think there are many other tables on this site.

posts/2018-08-07-refactoring-monads.md
```diff
@@ -0,0 +1,389 @@
...
+describing your function with a monadic for-comprehension. Hopefully
+you will find it easy to express the overall flow of logic this way.
+
+I suggest starting with the for-comprehension and writing pseudocode.
+Don't define the individual functions that comprise the steps of the
+for-comprehension until you have filled in the `yield` at the end.
+This is a great time to shuffle steps around and work out exactly what
+arguments are needed and when, and where they are coming from.
+
+Once the top level function and looks plausible, begin refactoring
```
uglyfigurine: Additional word `and` in `top level function and looks`

posts/2018-08-07-refactoring-monads.md
```diff
@@ -0,0 +1,391 @@
...
+implicit class SqlHelper(private val sc: StringContext) extends AnyVal {
+  def sql(args: Any*): Sql = Sql(args.mkString)
+}
+```
+
+I was recently cleaning up some Scala code I'd written a few months
+ago when I realized I had been structuring code in a very confusing
+way for a very long time. At work, we've been trying to combat the
+inevitable spaghetti of code that gets written by different authors at
+different times, as requirements inevitably change. We all know that
```
larsrh: duplication of "inevitable"

posts/2018-08-07-refactoring-monads.md
```diff
@@ -0,0 +1,391 @@
...
+understanding all three functions both individually and as a whole. To
+follow the logic, we must trace the functions like a recursive descent
+parser.
+
+### Refactoring with Monads
+Without throwing exceptions and catching them at the top, it's going
+to be hard to do substantially better than the "tail-refactoring"
+approach, unless we start to make use of the fact that we're working
+with `Try`, a data type that supports `flatMap`. More precisely, `Try`
+has a monad - recall that monads let us model computational effects
```
larsrh: "monad" → "monad instance"

posts/2018-08-07-refactoring-monads.md
```diff
@@ -0,0 +1,391 @@
...
+then use a for-comprehension to specify the sequence of operations:
+
+```tut:silent
+import scala.util.{Failure, Success, Try}
+
+def tryParse[A](s: String, parse: String => Option[A]): Try[A] = {
+  parse(s)
+    .map(Success(_))
+    .getOrElse(Failure(new ParseError(s)))
+}
```
larsrh: Wouldn't this be easier by having the parse functions return `Try` directly instead of `Option`? I was actually already wondering this above, where `ParseError` is defined but not immediately used.

posts/2018-08-07-refactoring-monads.md
```diff
@@ -0,0 +1,391 @@
...
+        parseDownloadType(row[Option[String]]("download_type"), id)
+          .map(EBook(id, title, author, _))
+    }
+  } yield book
+```
+
+It's less code, the functions are smaller, and the top-level function
+dictates the entire flow of control. No function takes more than 2
+arguments. These are testable, understandable functions. This version
+really shows the power of using Monads to sequence computation.
```
larsrh: "Monads" → "monads"

posts/2018-08-07-refactoring-monads.md
```diff
@@ -0,0 +1,391 @@
...
+
+Now we are truly making use of the fact that `Try` has a monad and not
```
larsrh: again, "monad instance"

posts/2018-08-07-refactoring-monads.md
mtomko: Yes, I think so.

We tend to use a return type of `Option` when writing something like `.fromString` because it's a little lighter weight, isn't opinionated as to whether a missing value is an error or not, and it's easy to convert to `Try` with an appropriate exception type later with adapters like `tryParse`. I did the same here out of habit.

I'll switch the `fromString` methods over to something that returns `Try` instead, so we can get rid of `tryParse` entirely. The example should still hold up and there'll be less code for people to grok.

posts/2018-08-07-refactoring-monads.md
mtomko: Thanks. I'm never sure about how to write that, but "has a monad instance" seems right. I'm not a category theory expert, but it always feels wrong to say "X is a monad" but rather "there is a monad for X". By the definition, it seems like a monad is really something that exists outside of, but in conjunction with whatever X is.

mtomko:
I've updated the pull request to address your suggestions. We still have the badly formatted table to deal with, and possibly should update the date of the blog post if we want it to reflect the publication date.

larsrh:
No worries, I'll deal with that.

posts/2018-08-07-refactoring-monads.md
larsrh: Perfect, thanks!

larsrh:
@mtomko Preview: https://deploy-preview-227--typelevel.netlify.com/blog/2018/08/07/refactoring-monads.html

Does that look alright to you?

larsrh:
Done! Will tweet about it shortly.

PR #228, opened 2018-08-25
Blog post: Http4s error handling using Cats Meow MTL

gvolpe:
Please have a look @oleg-py !

~I used `tut` for everything except for the last part because the dependency `meow-mtl` could not be resolved. If anybody has any idea of how to solve it (maybe adding a Resolver for tut?) I'd be glad to fix it.~ 

Feedback from @rossabaker would be greatly appreciated :)

### DIFF
--- a/posts/2018-08-25-http4s-error-handling-mtl.md
+++ b/posts/2018-08-25-http4s-error-handling-mtl.md
@@ -44,7 +44,7 @@ trait UserAlgebra[F[_]] {
 }
 ```
 
-And also and ADT of the possible errors that may arise. I'll explain later in this post why it extends `Throwable`.
+And also an ADT of the possible errors that may arise. I'll explain later in this post why it extends `Throwable`.
 
 ```tut:book:silent
 sealed trait UserError extends Throwable
@@ -194,7 +194,7 @@ I heard sometime ago about `Classy Optics` (Lenses, Prisms, etc) when I was lear
 
 Well first, let me give you a quick definition of `Lenses` and `Prisms`. In a few words we can define:
 
-- `Lenses` as getters and setters that compose making easy accessing fields of nested data structures.
+- `Lenses` as getters and setters that compose making the accessing of nested data structure's fields quite easy.
 - `Prisms` as first-class pattern matching that let us access branches of an ADT and that also compose.
 
 And the idea of `Classy Optics` is to "associate each type a typeclass full of optics for that type".
--- a/posts/2018-08-25-http4s-error-handling-mtl.md
+++ b/posts/2018-08-25-http4s-error-handling-mtl.md
@@ -45,10 +45,10 @@ trait UserAlgebra[F[_]] {
 }
 ```
 
-And also an ADT of the possible errors that may arise. I'll explain later in this post why it extends `Throwable`.
+And also an ADT of the possible errors that may arise. I'll explain later in this post why it extends `Exception`.
 
 ```tut:book:silent
-sealed trait UserError extends Throwable
+sealed trait UserError extends Exception
 case class UserAlreadyExists(username: String) extends UserError
 case class UserNotFound(username: String) extends UserError
 case class InvalidUserAge(age: Int) extends UserError
@@ -182,23 +182,23 @@ I heard sometime ago about Classy Optics (Lenses, Prisms, etc) when I was learni
 
 Well first, let me give you a quick definition of `Lens`es and `Prism`s. In a few words we can define:
 
-- `Lenses` as getters and setters that compose making the accessing of nested data structure's fields quite easy.
-- `Prisms` as first-class pattern matching that let us access branches of an ADT and that also compose.
+- `Lens`es as getters and setters that compose making the accessing of nested data structure's fields quite easy.
+- `Prism`s as first-class pattern matching that let us access branches of an ADT and that also compose.
 
 And `Classy Optics` as the idea of "associate with each type a typeclass full of optics for that type".
 
 ***So what am I talking about and how can these concepts help us solving the http error handling problem?***
 
-Remember that I defined the `UserError` ADT by extending `Throwable`?
+Remember that I defined the `UserError` ADT by extending `Exception`?
 
 ```scala
-sealed trait UserError extends Throwable
+sealed trait UserError extends Exception
 case class UserAlreadyExists(username: String) extends UserError
 case class UserNotFound(username: String) extends UserError
 case class InvalidUserAge(age: Int) extends UserError
 ```
 
-Well there's a reason! By making `UserError` a subtype of `Throwable` we can take advantage of `Prisms` by going back and forth in the types. See what I'm going yet?
+Well there's a reason! By making `UserError` a subtype of `Exception` (and by default of `Throwable`) we can take advantage of `Prisms` by going back and forth in the types. See what I'm going yet?
 
 `UserRoute` has a `Sync[F]` constraint, meaning that we have available a `MonadError[F, Throwable]` instance, but we would like to have `MonadError[F, UserError]` instead to leverage the Scala compiler. The caveat is that the error types need to be of the same family so we can derive a `Prism` that can navigate the errors types in one direction or another. But how do we derive it?
 
--- a/posts/2018-08-25-http4s-error-handling-mtl.md
+++ b/posts/2018-08-25-http4s-error-handling-mtl.md
@@ -20,6 +20,7 @@ tut:
     - io.circe::circe-core:0.10.0-M1
     - io.circe::circe-generic:0.10.0-M1
     - co.fs2::fs2-core:1.0.0-M1
+    - com.olegpy::meow-mtl:0.1.1
 
 ---
 
@@ -55,7 +56,7 @@ case class InvalidUserAge(age: Int) extends UserError
 
 ### User Interpreter
 
-And here we have a simple interpreter for our `UserAlgebra` for demonstration purposes so you can have an idea on how the logic would look like. In a real-life project an interpreter will more likely have some `Repository` as a dependency instead of using a `Ref`.
+And here we have a simple interpreter for our `UserAlgebra` for demonstration purposes so you can have an idea on how the logic would look like. In a real-life project an interpreter will more likely connect to a database instead of using an in-memory representaion based on `Ref`.
 
 ```tut:book:silent
 import cats.effect.Sync
@@ -98,45 +99,33 @@ object UserInterpreter {
 
 ### Http Routes
 
-Right before we define the `UserRoutes` we need to write some generic Json codecs to avoid repeating ourselves:
-
-```tut:book:silent
-import io.circe._
-import io.circe.generic.auto._
-import org.http4s._
-import org.http4s.circe._
-
-object JsonCodecs {
-  implicit def jsonDecoder[A <: Product: Decoder, F[_]: Sync]: EntityDecoder[F, A] = jsonOf[F, A]
-  implicit def jsonEncoder[A <: Product: Encoder, F[_]: Sync]: EntityEncoder[F, A] = jsonEncoderOf[F, A]
-}
-```
-
 The following implementation of `UserRoutes` applies the tagless final encoding and the concept of "abstracting over the effect type" where we do not commit to a particular effect until the edge of our application.
 
 ```tut:book:silent
+import io.circe.generic.auto._
+import org.http4s.{ HttpRoutes, Response }
+import org.http4s.circe._
+import org.http4s.circe.CirceEntityCodec._
 import org.http4s.dsl.Http4sDsl
 
 class UserRoutes[F[_]: Sync](userAlgebra: UserAlgebra[F]) extends Http4sDsl[F] {
-
-  import JsonCodecs._
 
   val routes: HttpRoutes[F] = HttpRoutes.of[F] {
 
     case GET -> Root / "users" / username =>
       userAlgebra.find(username).flatMap {
         case Some(user) => Ok(user)
-        case None => NotFound(username)
+        case None => NotFound()
       }
 
     case req @ POST -> Root / "users" =>
       req.as[User].flatMap { user =>
-        userAlgebra.save(user) *> Created(user.username)
+        userAlgebra.save(user) *> Created()
       }
 
     case req @ PUT -> Root / "users" / username =>
       req.as[UserUpdateAge].flatMap { userUpdate =>
-        userAlgebra.updateAge(username, userUpdate.age) *> Ok(username)
+        userAlgebra.updateAge(username, userUpdate.age) *> Ok()
       }
   }
 
@@ -154,28 +143,26 @@ Okay let's just go ahead and add some error handling to our http route by taking
 ```tut:book:silent
 class UserRoutesAlt[F[_]: Sync](userAlgebra: UserAlgebra[F]) extends Http4sDsl[F] {
 
-  import JsonCodecs._
-
   val routes: HttpRoutes[F] = HttpRoutes.of[F] {
 
     case GET -> Root / "users" / username =>
       userAlgebra.find(username).flatMap {
         case Some(user) => Ok(user)
-        case None => NotFound(username)
+        case None => NotFound()
       }
 
     case req @ POST -> Root / "users" =>
       req.as[User].flatMap { user =>
-        userAlgebra.save(user) *> Created(user.username)
-      }.handleErrorWith { // compiles without giving you "match non-exhaustive" error
-        case UserAlreadyExists(username) => Conflict(username)
+        userAlgebra.save(user) *> Created()
+      }.handleErrorWith {
+        case UserAlreadyExists(username) => Conflict()
       }
 
     case req @ PUT -> Root / "users" / username =>
       req.as[UserUpdateAge].flatMap { userUpdate =>
-        userAlgebra.updateAge(username, userUpdate.age) *> Ok(username)
-      }.handleErrorWith { // compiles without giving you "match non-exhaustive" error
-        case InvalidUserAge(age) => BadRequest(s"Invalid age $age")
+        userAlgebra.updateAge(username, userUpdate.age) *> Ok()
+      }.handleErrorWith {
+        case InvalidUserAge(age) => BadRequest()
       }
   }
 
@@ -190,14 +177,14 @@ So, what can we do about it?
 
 ### Next level MTL: Optics
 
-I heard sometime ago about `Classy Optics` (Lenses, Prisms, etc) when I was learning `Haskell` and watched [this amazing talk](https://www.youtube.com/watch?v=GZPup5Iuaqw) by George Wilson but I never got to use this concept in Scala until now!
+I heard sometime ago about Classy Optics (Lenses, Prisms, etc) when I was learning Haskell and watched [this amazing talk](https://www.youtube.com/watch?v=GZPup5Iuaqw) by George Wilson but I never got to use this concept in Scala until now!
 
-Well first, let me give you a quick definition of `Lenses` and `Prisms`. In a few words we can define:
+Well first, let me give you a quick definition of `Lens`es and `Prism`s. In a few words we can define:
 
 - `Lenses` as getters and setters that compose making the accessing of nested data structure's fields quite easy.
 - `Prisms` as first-class pattern matching that let us access branches of an ADT and that also compose.
 
-And the idea of `Classy Optics` is to "associate each type a typeclass full of optics for that type".
+And `Classy Optics` as the idea of "associate with each type a typeclass full of optics for that type".
 
 ***So what am I talking about and how can these concepts help us solving the http error handling problem?***
 
@@ -212,7 +199,7 @@ case class InvalidUserAge(age: Int) extends UserError
 
 Well there's a reason! By making `UserError` a subtype of `Throwable` we can take advantage of `Prisms` by going back and forth in the types. See what I'm going yet?
 
-`UserRoute` has a `Sync[F]` constraint meaning that we have available a `MonadError[F, Throwable]` instance but we would like to have `MonadError[F, UserError]` instead to leverage the Scala compiler. There's a caveat and is that the error types need to be of the same family so we can derive a `Prism` that can navigate the errors types in one direction or another. But how do we derive it?
+`UserRoute` has a `Sync[F]` constraint, meaning that we have available a `MonadError[F, Throwable]` instance, but we would like to have `MonadError[F, UserError]` instead to leverage the Scala compiler. The caveat is that the error types need to be of the same family so we can derive a `Prism` that can navigate the errors types in one direction or another. But how do we derive it?
 
 #### Cats Meow MTL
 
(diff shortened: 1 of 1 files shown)


posts/2018-08-25-http4s-error-handling-mtl.md
```diff
@@ -0,0 +1,327 @@
...
+case class UserUpdateAge(age: Int)
+
+trait UserAlgebra[F[_]] {
+  def find(username: String): F[Option[User]]
+  def save(user: User): F[Unit]
+  def updateAge(username: String, age: Int): F[Unit]
+}
+```
+
+And also and ADT of the possible errors that may arise. I'll explain later in this post why it extends `Throwable`.
```
Avasil: typo

posts/2018-08-25-http4s-error-handling-mtl.md
```diff
@@ -0,0 +1,327 @@
...
+```tut:book:silent
+sealed trait UserError extends Throwable
+case class UserAlreadyExists(username: String) extends UserError
+case class UserNotFound(username: String) extends UserError
+case class InvalidUserAge(age: Int) extends UserError
+```
+
+### User Interpreter
+
+And here we have a simple interpreter for our `UserAlgebra` for demonstration purposes so you can have an idea on how the logic would look like. In a real-life project an interpreter will more likely have some `Repository` as a dependency instead of using a `Ref`.
```
rossabaker: `Repository` is a term from my Java/DDD days.  Is this a well known concept to Scala developers?

I might say something like "more likely connect to a database.  Here, we use an in-memory implementation based on a `Ref`"

posts/2018-08-25-http4s-error-handling-mtl.md
```diff
@@ -0,0 +1,327 @@
...
+import cats.effect.Sync
+import cats.effect.concurrent.Ref
+import cats.syntax.all._
+
+object UserInterpreter {
+
+  def create[F[_]](implicit F: Sync[F]): F[UserAlgebra[F]] =
+    Ref.of[F, Map[String, User]](Map.empty).map { state =>
+      new UserAlgebra[F] {
+        private def validateAge(age: Int): F[Unit] =
```
rossabaker: I normally think of validators as returning the validated value on the happy path and probably would have returned an `F[Int]`.  But since you're not transforming it or lifting it into a newtype, I guess it doesn't really matter.

posts/2018-08-25-http4s-error-handling-mtl.md
```diff
@@ -0,0 +1,327 @@
...
+
+Right before we define the `UserRoutes` we need to write some generic Json codecs to avoid repeating ourselves:
+
+```tut:book:silent
+import io.circe._
+import io.circe.generic.auto._
+import org.http4s._
+import org.http4s.circe._
+
+object JsonCodecs {
```
rossabaker: This exists as [CirceEntityCodecs](https://github.com/http4s/http4s/blob/v0.19.0-M1/circe/src/main/scala/org/http4s/circe/CirceEntityCodec.scala) now.

posts/2018-08-25-http4s-error-handling-mtl.md
```diff
@@ -0,0 +1,327 @@
...
+    case GET -> Root / "users" / username =>
+      userAlgebra.find(username).flatMap {
+        case Some(user) => Ok(user)
+        case None => NotFound(username)
+      }
+
+    case req @ POST -> Root / "users" =>
+      req.as[User].flatMap { user =>
+        userAlgebra.save(user) *> Created(user.username)
+      }.handleErrorWith { // compiles without giving you "match non-exhaustive" error
```
rossabaker: This comment is confusing: is there any reason a user should expect a warning?

posts/2018-08-25-http4s-error-handling-mtl.md
```diff
@@ -0,0 +1,327 @@
...
+
+The problem is that our constraint of type `Sync` from `cats-effect` has a `MonadError` instance with its type error fixed as `Throwable`. So the compiler can't help us here since this type is too generic. And we can't add a constraint for `MonadError[F, UserError]` because we would get an "ambigous implicits" error with two instances of `MonadError` in scope.
+
+So, what can we do about it?
+
+### Next level MTL: Optics
+
+I heard sometime ago about `Classy Optics` (Lenses, Prisms, etc) when I was learning `Haskell` and watched [this amazing talk](https://www.youtube.com/watch?v=GZPup5Iuaqw) by George Wilson but I never got to use this concept in Scala until now!
+
+Well first, let me give you a quick definition of `Lenses` and `Prisms`. In a few words we can define:
```
rossabaker: I generally like to put the grammatical inflection outside the coded type: `Lens`es, `Prism`s.

posts/2018-08-25-http4s-error-handling-mtl.md
```diff
@@ -0,0 +1,327 @@
...
+
+Now we can say this implementation is quite elegant! We are handling and mapping business errors to the according http response and our code compiles without any warning whatsoever. But wait... We are not handling the `UserNotFound` error and the compiler didn't tell us about it! That's not cool and we as functional programmers believe in types because we can know what a function might do just by looking at the types but here it seems we hit the wall.
+
+The problem is that our constraint of type `Sync` from `cats-effect` has a `MonadError` instance with its type error fixed as `Throwable`. So the compiler can't help us here since this type is too generic. And we can't add a constraint for `MonadError[F, UserError]` because we would get an "ambigous implicits" error with two instances of `MonadError` in scope.
+
+So, what can we do about it?
+
+### Next level MTL: Optics
+
+I heard sometime ago about `Classy Optics` (Lenses, Prisms, etc) when I was learning `Haskell` and watched [this amazing talk](https://www.youtube.com/watch?v=GZPup5Iuaqw) by George Wilson but I never got to use this concept in Scala until now!
```
rossabaker: Haskell isn't a code literal, it's just a proper noun.  Classy Optics is just a noun, too.

posts/2018-08-25-http4s-error-handling-mtl.md
```diff
@@ -0,0 +1,327 @@
...
+
+Well first, let me give you a quick definition of `Lenses` and `Prisms`. In a few words we can define:
+
+- `Lenses` as getters and setters that compose making the accessing of nested data structure's fields quite easy.
+- `Prisms` as first-class pattern matching that let us access branches of an ADT and that also compose.
+
+And the idea of `Classy Optics` is to "associate each type a typeclass full of optics for that type".
```
rossabaker: I think there's a missing "with" here.  And "associating each type with a typeclass" sounds funny to me.  Maybe "provide instances of a typeclass"?

posts/2018-08-25-http4s-error-handling-mtl.md
```diff
@@ -0,0 +1,327 @@
...
+```scala
+sealed trait UserError extends Throwable
+case class UserAlreadyExists(username: String) extends UserError
+case class UserNotFound(username: String) extends UserError
+case class InvalidUserAge(age: Int) extends UserError
+```
+
+Well there's a reason! By making `UserError` a subtype of `Throwable` we can take advantage of `Prisms` by going back and forth in the types. See what I'm going yet?
+
+`UserRoute` has a `Sync[F]` constraint meaning that we have available a `MonadError[F, Throwable]` instance but we would like to have `MonadError[F, UserError]` instead to leverage the Scala compiler. There's a caveat and is that the error types need to be of the same family so we can derive a `Prism` that can navigate the errors types in one direction or another. But how do we derive it?
```
rossabaker: "The caveat is that the..."

Commas for the clause: "constraint, meaning that ... instance, but"

posts/2018-08-25-http4s-error-handling-mtl.md
```diff
@@ -0,0 +1,327 @@
...
+
+#### Cats Meow MTL
+
+Fortunately our friend [Oleg Pyzhcov](https://twitter.com/oleg_pyzhcov) has created this great library named [meow-mtl](https://github.com/oleg-py/meow-mtl) that makes heavy use of [Shapeless](https://github.com/milessabin/shapeless) in order to derive `Lenses` and `Prisms` and it provides instances for some `cats-effect` compatible datatypes.
+
+And two of the supported typeclasses are `ApplicativeError` and `MonadError` as long as the error type is a subtype of `Throwable` to make it compatible with `cats-effect`. So we can do something like this:
+
+```scala
```
rossabaker: Any reason this doesn't tut?

posts/2018-08-25-http4s-error-handling-mtl.md
```diff
@@ -0,0 +1,327 @@
...
+def customHandle[F[_]: MonadError[?[_], CustomError], A](f: F[A], fallback: F[A]): F[A] =
+  f.handleErrorWith(_ => fallback)
+
+val io: IO[Int] = ???
+customHandle(io, IO.pure(123))
+```
+
+#### Generalizing Http Error Handling
+
+Now back to our use case we can't have a `MonadError[F, UserError]` constraint because there's already a `MonadError[F, Throwable]` in scope given our `Sync[F]` constraint. But it turns out we can make this work if we also abstract over the error handling by introducing an `HttpErrorHandler` algebra where the error type is a subtype of `Throwable`.
```
rossabaker: Run-on sentence: "use case.  We can't"

posts/2018-08-25-http4s-error-handling-mtl.md
```diff
@@ -0,0 +1,327 @@
...
+  import JsonCodecs._
+  import syntax._
+
+  val routes: HttpRoutes[F] = HttpRoutes.of[F] {
+
+    case GET -> Root / "users" / username =>
+      userAlgebra.find(username).flatMap {
+        case Some(user) => Ok(user)
+        case None => NotFound(username)
+      }.handleHttpErrorResponse
```
rossabaker: This gets repeated for every route.  Can we go one step further and put `handleHttpErrorResponse` on `routes` itself?  It compiles in my head...

posts/2018-08-25-http4s-error-handling-mtl.md
```diff
@@ -0,0 +1,327 @@
...
+
+UserInterpreter.create[IO].flatMap { UserAlgebra =>
+  routes = new UserRoutesMTL[IO](UserAlgebra)
+  IO.unit // pretend this is the rest of your program
+}
+```
+
+### Final thoughts
+
+This is such an exciting time to be writing pure functional programming in Scala! The Typelevel ecosystem is getting richer and more mature having an amazing set of libraries to solve business problems in an elegant and purely functional way.
```
rossabaker: Comma after mature

posts/2018-08-25-http4s-error-handling-mtl.md
```diff
@@ -0,0 +1,327 @@
...
+
+I hope you have enjoyed this post and please do let me know if you know of better ways to solve this problem in the comments!
+
+And last but not least I would like to thank all the friendly folks I hang out with in the `cats-effect`, `cats`, `fs2` and `http4s` Gitter channels for all the time and effort they put (*for free*) into making this community an amazing space.
```
rossabaker: And many thanks to you for your contributions in all these spaces.  We :heart: good documenters.

posts/2018-08-25-http4s-error-handling-mtl.md
```diff
@@ -0,0 +1,327 @@
...
+  def find(username: String): F[Option[User]]
+  def save(user: User): F[Unit]
+  def updateAge(username: String, age: Int): F[Unit]
+}
+```
+
+And also an ADT of the possible errors that may arise. I'll explain later in this post why it extends `Throwable`.
+
+```tut:book:silent
+sealed trait UserError extends Throwable
```
rossabaker: Extending `Throwable` directly makes me cringe as an old Java developer.  I think extending `Exception` is much more idiomatic, although then you'd have to explain that `Exception` is a `Throwable` and distracts from your point.

I'll leave this one to your discretion.

larsrh:
> I used tut for everything except for the last part because the dependency meow-mtl could not be resolved. If anybody has any idea of how to solve it (maybe adding a Resolver for tut?) I'd be glad to fix it.

@gvolpe It resolves fine from my end: `com.olegpy::meow-mtl:0.1.1`.

posts/2018-08-25-http4s-error-handling-mtl.md
gvolpe: I think I get your point but since it's just a simple interpreter for the user to have an idea of how the business logic would look like I didn't want to get into `validation`. I think leaving it as it is wouldn't be that bad since I really want the users to focus on the error handling topic instead.

posts/2018-08-25-http4s-error-handling-mtl.md
gvolpe: Amazing! Didn't know about it :)

posts/2018-08-25-http4s-error-handling-mtl.md
gvolpe: I literally tried to take George's definition from [his slides](https://github.com/gwils/next-level-mtl-with-classy-optics/blob/master/Slides.pdf) but I even failed doing that. Hope the correction is fine!

posts/2018-08-25-http4s-error-handling-mtl.md
gvolpe: Sounds great! I'll try to get it compiling and working :)

gvolpe:
Thanks a lot for your review @rossabaker !

One thing that happened after removing my custom `json` codecs and adding the suggested `CirceEntityCodecs` is that now I get this error when trying to return a response of type `String`. How can I solve this?

```
// <console>:41: error: ambiguous implicit values:
//  both method stringEncoder in trait EntityEncoderInstances of type [F[_]](implicit charset: org.http4s.Charset)org.http4s.EntityEncoder[F,String]
//  and method circeEntityEncoder in trait CirceEntityEncoder of type [F[_], A](implicit evidence$1: cats.Applicative[F], implicit evidence$2: io.circe.Encoder[A])org.http4s.EntityEncoder[F,A]
//  match expected type org.http4s.EntityEncoder[F,String]
// Error occurred in an application involving default arguments.
//                case None => NotFound(username)
```

Thanks @larsrh ! I got it compiling now.... Might have been a connection issue at that time I guess.

oleg-py:
@gvolpe great post, thanks for highlighting meow-mtl 👍

gvolpe:
Hope it becomes part of `cats-mtl` soon @oleg-py :)

rossabaker:
Oh, right.  This is why I don't like `CirceEntityCodecs`: it introduces incoherent instance, which are hard to reason about, especially in tut scopes.  This is why I always use `.asJson` and then rely on the `EntiityEncoder[F, Json]` you get from `org.http4s.circe._`.  It's still not ideal, because it's an _orphan_ instance, but at least it's unique.

posts/2018-08-25-http4s-error-handling-mtl.md
gvolpe: @rossabaker can you expand on this idea? I'm now trying something out but nothing seems correct. This is what I've come up with so far (compiling):

```scala
trait HttpErrorHandler[F[_], E <: Throwable] {
  def handle(fa: OptionT[F, Response[F]]): OptionT[F, Response[F]]
}

object HttpErrorHandler {
  def apply[F[_], E <: Throwable](implicit ev: HttpErrorHandler[F, E]) = ev
}

object syntax {
  implicit class HttpErrorHandlerOp[F[_]: Functor, E <: Throwable](routes: HttpRoutes[F]) {
    def handleHttpErrorResponse(implicit H: HttpErrorHandler[F, E]): HttpRoutes[F] = {
//      import org.http4s.implicits._
//      val app: HttpApp[F] = routes.orNotFound // adding orNotFound here does not seem correct
//      app.mapF(fa => OptionT.liftF(H.handle(fa)))
        routes.mapF(H.handleOpt)
    }
  }
}

class UserRoutesMTL[F[_]: Sync](userAlgebra: UserAlgebra[F])
                               (implicit H: HttpErrorHandler[F, UserError]) extends Http4sDsl[F] {

  private val httpRoutes: HttpRoutes[F] = HttpRoutes.of[F] { ... }

  val routes: HttpRoutes[F] = httpRoutes.handleHttpErrorResponse
}

class UserHttpErrorHandler[F[_]: MonadError[?[_], UserError]] extends HttpErrorHandler[F, UserError] with Http4sDsl[F] {
  override def handleOpt(fa: OptionT[F, Response[F]]) = {
    OptionT {
      fa.value.handleErrorWith {
        case InvalidUserAge(age) => BadRequest(s"Invalid age $age".asJson).map(Option(_)) // argh
        case UserAlreadyExists(username) => Conflict(username.asJson).map(Option(_))
        case UserNotFound(username) => NotFound(username.asJson).map(Option(_))
      }
    }
  }
}
```

gvolpe: Well this one seems a bit more acceptable but I'd like to hear how you would do it :smile: 

```scala

trait HttpErrorHandler[F[_], E <: Throwable] {
 def handle(routes: HttpRoutes[F]): HttpRoutes[F]
}

object HttpErrorHandler {
  def apply[F[_], E <: Throwable](implicit ev: HttpErrorHandler[F, E]) = ev
}

object syntax {
  implicit class HttpErrorHandlerOp[F[_]: Functor, E <: Throwable](routes: HttpRoutes[F]) {
    def handleHttpErrorResponse(implicit H: HttpErrorHandler[F, E]): HttpRoutes[F] = {
      H.handle(routes)
    }
  }
}

class UserHttpErrorHandler[F[_]: MonadError[?[_], UserError]] extends HttpErrorHandler[F, UserError] with Http4sDsl[F] {
  override def handle(routes: HttpRoutes[F]): HttpRoutes[F] = {
    val handler: UserError => F[Response[F]] = {
      case InvalidUserAge(age) => BadRequest(s"Invalid age $age".asJson)
      case UserAlreadyExists(username) => Conflict(username.asJson)
      case UserNotFound(username) => NotFound(username.asJson)
    }
    
    Kleisli { req =>
      OptionT {
        routes.run(req).value.handleErrorWith(e => handler(e).map(Option(_)))
      }
    }
  }
}
```

gvolpe: And going one step further I can extract that Kleisli stuff out in an abstract class:

```scala
trait HttpErrorHandler[F[_], E <: Throwable] {
  def handle: E => F[Response[F]]
  def routesHandler: HttpRoutes[F] => HttpRoutes[F]
}

abstract class DefaultHttpErrorHandler[F[_]: MonadError[?[_], E], E <: Throwable] extends HttpErrorHandler[F, E] {
  override def routesHandler: HttpRoutes[F] => HttpRoutes[F] = { routes =>
    Kleisli { req =>
      OptionT {
        routes.run(req).value.handleErrorWith(e => handle(e).map(Option(_)))
      }
    }
  }
}

object HttpErrorHandler {
  def apply[F[_], E <: Throwable](implicit ev: HttpErrorHandler[F, E]) = ev
}

object syntax {
  implicit class HttpErrorHandlerOps[F[_], E <: Throwable](routes: HttpRoutes[F]) {
    def handleHttpErrorResponse(implicit H: HttpErrorHandler[F, E]): HttpRoutes[F] = {
      H.routesHandler(routes)
    }
  }
}

class UserHttpErrorHandler[F[_]: MonadError[?[_], UserError]] extends DefaultHttpErrorHandler[F, UserError] with Http4sDsl[F] {
  override def handle: UserError => F[Response[F]] = {
    case InvalidUserAge(age) => BadRequest(s"Invalid age $age".asJson)
    case UserAlreadyExists(username) => Conflict(username.asJson)
    case UserNotFound(username) => NotFound(username.asJson)
  }
}
```

Looking forward to your feedback!!! :smile:

gvolpe:
@rossabaker I left a few comments in an outdated comment so I'm not sure whether you got notified?

posts/2018-08-25-http4s-error-handling-mtl.md
gvolpe: And the last one (woke up with some clearer ideas :smile: )

```scala
trait HttpErrorHandler[F[_], E <: Throwable] {
  def handle(routes: HttpRoutes[F]): HttpRoutes[F]
}

object RoutesHttpErrorHandler {
  def apply[F[_]: MonadError[?[_], E], E <: Throwable](routes: HttpRoutes[F])(handler: E => F[Response[F]]): HttpRoutes[F] = {
    Kleisli { req =>
      OptionT {
        routes.run(req).value.handleErrorWith(e => handler(e).map(Option(_)))
      }
    }
  }
}

object HttpErrorHandler {
  def apply[F[_], E <: Throwable](implicit ev: HttpErrorHandler[F, E]) = ev
}

object syntax {
  implicit class HttpErrorHandlerOp[F[_], E <: Throwable](routes: HttpRoutes[F]) {
    def handleHttpErrorResponse(implicit H: HttpErrorHandler[F, E]): HttpRoutes[F] = {
      H.handle(routes)
    }
  }
}

class UserHttpErrorHandler[F[_]: MonadError[?[_], UserError]] extends HttpErrorHandler[F, UserError] with Http4sDsl[F] {
  private val handler: UserError => F[Response[F]] = {
    case InvalidUserAge(age) => BadRequest(s"Invalid age $age".asJson)
    case UserAlreadyExists(username) => Conflict(username.asJson)
    case UserNotFound(username) => NotFound(username.asJson)
  }

  override def handle(routes: HttpRoutes[F]): HttpRoutes[F] =
    RoutesHttpErrorHandler(routes)(handler)
}
```

posts/2018-08-25-http4s-error-handling-mtl.md
```diff
@@ -0,0 +1,333 @@
...
+We are basically delegating the error handling (AKA mapping business errors to appropiate http responses) to a specific algebra by making use of the syntax we have previously introduced.
+
+We also need an implementation for this algebra in order to handle errors of type `UserError` but first we can introduce a `RoutesHttpErrorHandler` object that encapsulates the repetitive task of handling errors given an `HttpRoutes[F]` to generalize things a bit more :)
+
+```tut:book:silent
+import cats.data.{Kleisli, OptionT}
+
+object RoutesHttpErrorHandler {
+  def apply[F[_], E <: Throwable](routes: HttpRoutes[F])(handler: E => F[Response[F]])(implicit M: MonadError[F, E]): HttpRoutes[F] =
+    Kleisli { req: Request[F] =>
```
gvolpe: I can't get `tut` to like this, any compiler flag that needs to be set?

gvolpe:
I finally went with my latest proposal for generalizing error handling at the `HttpRoutes[F]` level. I think I'm closer now but still I'll wait for @rossabaker 's final words.

posts/2018-08-25-http4s-error-handling-mtl.md
rossabaker: I don't see that `-Ypartial-unification` is on here, which often helps with inference puzzlers.  Though setting it here in the SBT repl didn't seem to make a difference.  It works if you're very explicit about types, and we don't have kind-projector in this blog environment, which makes it worse:

```scala
    Kleisli[({ type T[A] = OptionT[F, A] })#T, Request[F], Response[F]] { req: Request[F] =>
      OptionT {
        routes.run(req).value.handleErrorWith { e => handler(e).map(Option(_)) }
      }
    }
```

There's also an `HttpRoutes.apply`, but that takes a `Sync` constraint, whose `MonadError` fights with `M`:

```scala
  def apply2[F[_], E <: Throwable](routes: HttpRoutes[F])(handler: E => F[Response[F]])(implicit F: Sync[F], M: MonadError[F, E]): HttpRoutes[F] =
    HttpRoutes { req: Request[F] =>
      OptionT {
        M.handleErrorWith(routes.run(req).value) { e => M.map(handler(e))(Option(_)) }
      }
    }
```

gvolpe: Thanks for investigating Ross.

I guess that if I have to sacrifice readability I'd rather not `tut`... Surely that was a question I wanted to ask, `kind-projector` is not supported then?

larsrh: All of this can be implemented (both the flags and the compiler plugin). It's just that somebody's got to do it [here](https://github.com/typelevel/typelevel.github.com/tree/development/src/main/scala).

gvolpe: @larsrh interesting. If you or somebody else can provide any guidance I can do it as part of this PR. For example how do I pass `scalacOptions` settings for `tut`?

gvolpe:
@larsrh @rossabaker are you okay with the content of the post?

I'm thinking that it might be better to just skip `tut` on the problematic blocks and once we can fix how `tut` runs I can PR a fix for it. What do you think?

rossabaker:
Yes, I think the content is good.

larsrh:
> I'm thinking that it might be better to just skip tut on the problematic blocks and once we can fix how tut runs I can PR a fix for it. What do you think?

Sounds good to me. I think this is good to merge once the build passes.

gvolpe:
Awesome thanks a lot @larsrh  and @rossabaker for your fantastic review :smiley: 

For the record I started investigating and looking into `TutMain` to see if I can do something about it :)

larsrh:
@gvolpe Thank _you_ for the contribution!

PR #229, opened 2018-09-02
[Corrections] - Error handling MTL blog post

gvolpe:
- Removing syntax: I realized it was no longer necessary since now the function is only called once at the routes level whereas before it was called multiple times per response (hence it made sense having syntax).
- Relaxing constraint of `RoutesHttpErrorHandler.apply` to `ApplicativeError`.


larsrh:
@gvolpe I gave you push access to this repository. Merge at will!

gvolpe:
Awesome, thanks @larsrh !

PR #230, opened 2018-09-04
Add Chain Blog post

LukaJCB:
We can only merge this when `Cats` 1.3.0 is released, I'll also need to do some proofreading.

### DIFF
--- a/posts/2018-09-04-chain-replacing-the-list-monoid.md
+++ b/posts/2018-09-04-chain-replacing-the-list-monoid.md
@@ -12,7 +12,7 @@ meta:
 `List` is a great data type, it is very simple and easy to understand.
 It has very low overhead for the most important functions such as `fold` and `map` and also supports prepending a single element in constant time.
 
-Traversing a data structure with something like `Writer[List[Log], A]` or `ValidatedNel[Error, A]`, is really powerful and allows us to very precisely specify what kind of iteration we want to do while remaining very succint.
+Traversing a data structure with something like `Writer[List[Log], A]` or `ValidatedNel[Error, A]` is  powerful and allows us to precisely specify what kind of iteration we want to do while remaining succint.
 However, in terms of efficiency it's a whole different story unfortunately.
 That is because both of these traversals make use of the `List` monoid (or the `NonEmptyList` semigroup), which by the nature of `List` is very inefficient.
 If you use `traverse` with a data structure with `n` elements and `Writer` or `Validated` as the `Applicative` type, you will end up with a runtime of `O(n^2)`.
@@ -25,11 +25,11 @@ Well, `Vector` has its own problems and in this case it's unfortunately not all 
 Because of this, it's now time to welcome a new data structure to Cats.
 Meet `Chain` and it's non-empty counterpart `NonEmptyChain`. 
 
-Available in the newest Cats 1.3.0 release, `Chain` is somewhat derived from what used to be `fs2.Catenable` and Erik Osheim's [Chain](https://github.com/non/chain ) library.
-Similar to `List` it is also a very simple data structure, but unlike `List` it supports both constant time `append` and `prepend`.
+Available in the newest Cats 1.3.0 release, `Chain` evolved from what used to be `fs2.Catenable` and Erik Osheim's [Chain](https://github.com/non/chain ) library.
+Similar to `List` it is also a very simple data structure, but unlike `List` it supports both constant O(1) time `append` and `prepend`.
 This makes its `Monoid` instance super performant and a much better fit for usage with `Validated`,`Writer`, `Ior` or `Const`.
 
-To utilize this, we've added a bunch of shorthands in Cats 1.3 that previously used `NonEmptyList` to use `NonEmptyChain` instead. These include type aliases like `ValidatedNec` or `IorNec` as well as helper functions like `groupByNec` or `Validated.invalidNec`.
+To utilize this, we've added a bunch of `NonEmptyChain` shorthands in Cats 1.3 that mirror those that used `NonEmptyList` in earlier versions. These include type aliases like `ValidatedNec` or `IorNec` as well as helper functions like `groupByNec` or `Validated.invalidNec`.
 We hope that these make it easy for you to upgrade to the more efficient data structure and enjoy those benefits as soon as possible.
 
 To get a good idea of the performance improvements, here are some benchmarks that test monoidal append to convince you of its efficiency (higher score is better):
--- a/posts/2018-09-04-chain-replacing-the-list-monoid.md
+++ b/posts/2018-09-04-chain-replacing-the-list-monoid.md
@@ -1,6 +1,6 @@
 ---
 layout: post
-title: PSA – Don't use the List monoid
+title: Chain – Replacing the List Monoid
 
 meta:
   nav: blog
--- a/posts/2018-09-04-chain-replacing-the-list-monoid.md
+++ b/posts/2018-09-04-chain-replacing-the-list-monoid.md
@@ -20,19 +20,19 @@ This is because, with `List`, appending a single element requires iterating over
 
 So `List` isn't all that great for this use case, so let's use `Vector` or `NonEmptyVector` instead, right?
 
-Well, `Vector` has its own problems and in this case it's unfortunately not all that faster than `List` at all. You can check [this blog post](http://www.lihaoyi.com/post/BenchmarkingScalaCollections.html#vectors-are-ok) by Li Haoyi for some deeper insight into `Vector`s issues.
+Well, `Vector` has its own problems and in this case it's unfortunately not that much faster than `List` at all. You can check [this blog post](http://www.lihaoyi.com/post/BenchmarkingScalaCollections.html#vectors-are-ok) by Li Haoyi for some deeper insight into `Vector`s issues.
 
 Because of this, it's now time to welcome a new data structure to Cats.
-Meet `Chain` and it's non-empty counterpart `NonEmptyChain`. 
+Meet `Chain` and it's non-empty counterpart, `NonEmptyChain`.
 
 Available in the newest Cats 1.3.0 release, `Chain` evolved from what used to be `fs2.Catenable` and Erik Osheim's [Chain](https://github.com/non/chain ) library.
-Similar to `List` it is also a very simple data structure, but unlike `List` it supports both constant O(1) time `append` and `prepend`.
+Similar to `List`, it is also a very simple data structure, but unlike `List` it supports both constant O(1) time `append` and `prepend`.
 This makes its `Monoid` instance super performant and a much better fit for usage with `Validated`,`Writer`, `Ior` or `Const`.
 
 To utilize this, we've added a bunch of `NonEmptyChain` shorthands in Cats 1.3 that mirror those that used `NonEmptyList` in earlier versions. These include type aliases like `ValidatedNec` or `IorNec` as well as helper functions like `groupByNec` or `Validated.invalidNec`.
 We hope that these make it easy for you to upgrade to the more efficient data structure and enjoy those benefits as soon as possible.
 
-To get a good idea of the performance improvements, here are some benchmarks that test monoidal append to convince you of its efficiency (higher score is better):
+To get a good idea of the performance improvements, here are some benchmarks that test monoidal append (higher score is better):
 
 ```
 [info] Benchmark                                  Mode  Cnt   Score   Error  Units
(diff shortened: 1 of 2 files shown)


posts/2018-09-04-do-not-use-list-monoid.md
```diff
@@ -0,0 +1,66 @@
+---
+layout: post
+title: PSA – Don't use the List monoid
```
larsrh: I would favour a title that sounds less like "List considered harmful" ☺

posts/2018-09-04-do-not-use-list-monoid.md
```diff
@@ -0,0 +1,66 @@
...
+
+meta:
+  nav: blog
+  author: lukajcb
+  pygments: true
+
+---
+
+`List` is a great data type, it is very simple and easy to understand.
+It has very low overhead for the most important functions such as `fold` and `map` and also supports prepending in constant time.
```
larsrh: "prepending _a single element_"

posts/2018-09-04-do-not-use-list-monoid.md
```diff
@@ -0,0 +1,66 @@
...
+
+Traversing a data structure with something like `Writer[List[Log], A]` or `ValidatedNel[Error, A]`, is really powerful and allows us to very precisely specify what kind of iteration we want to do while remaining very succint.
```
larsrh: "very"s sound superfluous

posts/2018-09-04-do-not-use-list-monoid.md
```diff
@@ -0,0 +1,66 @@
...
+However, in terms of efficiency it's a whole different story unfortunately.
+That is because both of these traversals make use of the `List` monoid (or the `NonEmptyList` semigroup), which by the nature of `List` is very ineffecient.
```
larsrh: typo: "inefficient"

posts/2018-09-04-do-not-use-list-monoid.md
```diff
@@ -0,0 +1,66 @@
...
+If you use `traverse` with a data structure with `n` elements and `Writer` or `Validated` as the `Applicative` type, you will end up with a runtime of `O(n^2)`.
+This is because, with `List`, appending a single element requires iterating over the entire data structure and therefore takes linear time.
+
+So `List` isn't all that great for this use case, so let's use `Vector` or `NonEmptyVector` instead, right?
+
+Well, `Vector` has its own problems and in this case it's unfortunately not all that faster than `List` at all. You can check [this blog post](http://www.lihaoyi.com/post/BenchmarkingScalaCollections.html#vectors-are-ok) by Li Haoyi for some deeper insight into `Vector`s issues.
```
larsrh: duplication: "not all that … at all"

posts/2018-09-04-do-not-use-list-monoid.md
```diff
@@ -0,0 +1,66 @@
...
+
+Because of this, it's now time to welcome a new data structure to Cats.
+Meet `Chain` and it's non-empty counterpart `NonEmptyChain`. 
+
+Available in the newest Cats 1.3.0 release, `Chain` is somewhat derived from what used to be `fs2.Catenable` and Erik Osheim's `Chain` library.
```
larsrh: Links, maybe?

posts/2018-09-04-do-not-use-list-monoid.md
```diff
@@ -0,0 +1,66 @@
...
+Similar to `List` it is also a very simple data structure, but unlike `List` it supports both constant time `append` and `prepend`.
+This makes it's `Monoid` instance super performant and a much better fit for usage with `Validated`,`Writer`, `Ior` or `Const`.
```
larsrh: typo: "its"

posts/2018-09-04-do-not-use-list-monoid.md
```diff
@@ -0,0 +1,66 @@
...
+
+```
+[info] Benchmark                                  Mode  Cnt   Score   Error  Units
+[info] CollectionMonoidBench.accumulateChain     thrpt   20  51.911 ± 7.453  ops/s
+[info] CollectionMonoidBench.accumulateList      thrpt   20   6.973 ± 0.781  ops/s
+[info] CollectionMonoidBench.accumulateVector    thrpt   20   6.304 ± 0.129  ops/s
+```
+
+As you can see accumulating things with `Chain` is more than 7 times faster than `List` and over 8 times faster than `Vector`.
+So appending seems to be a lot more performant than the standard library collections, but what about operations like `map` or `fold`.
```
larsrh: This is a question, but it ends with a full stop.

posts/2018-09-04-do-not-use-list-monoid.md
```diff
@@ -0,0 +1,66 @@
...
+We hope that these make it easy for you to upgrade to the more efficient data structure and enjoy those benefits as soon as possible.
+
+To get a good idea of the performance improvements, here are some benchmarks that test monoidal append to convince you of its efficiency.
+
+```
+[info] Benchmark                                  Mode  Cnt   Score   Error  Units
+[info] CollectionMonoidBench.accumulateChain     thrpt   20  51.911 ± 7.453  ops/s
+[info] CollectionMonoidBench.accumulateList      thrpt   20   6.973 ± 0.781  ops/s
+[info] CollectionMonoidBench.accumulateVector    thrpt   20   6.304 ± 0.129  ops/s
+```
```
larsrh: For clarification: higher score is better?

posts/2018-09-04-do-not-use-list-monoid.md
```diff
@@ -0,0 +1,66 @@
...
+[info] ChainBench.mapLargeVector          thrpt   20         61.374 ±       2.004  ops/s
+```
+
+While not as dominant, `Chain` can still hold its ground fairly well. 
+It won't have the random access performance of something like `Vector`, but in almost all other cases, `Chain` seems to outperform it quite handily.
+So if you don't perform a lot of random access on your data structure (I certainly don't), then you should be fine using `Chain` extensively instead.
+
+So next time you write any code that uses `List` or `Vector` as a `Monoid`, be sure to use `Chain` instead!
+
+The whole code for `Chain` and `NonEmptyChain` can be found [here](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/data/Chain.scala) and [here](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/data/NonEmptyChain.scala).
```
larsrh: Use permalinks (e.g. with the tag) here.

LukaJCB: I'm not sure what you mean by this

larsrh: These links:
https://github.com/typelevel/cats/blob/v1.3.0/core/src/main/scala/cats/data/Chain.scala
https://github.com/typelevel/cats/blob/v1.3.0/core/src/main/scala/cats/data/NonEmptyChain.scala

larsrh: (As opposed to links to `master`, these ones will still work even after someone moves around files.)

LukaJCB: Right, of course, sorry and thanks :smile:

posts/2018-09-04-do-not-use-list-monoid.md
LukaJCB: Any suggestions? :)

larsrh: Maybe "Chains: A more efficient data structure for concatenation"?

LukaJCB: Went with a different name that's still somewhat click-baity, but conveys the message well IMO :)

posts/2018-09-04-chain-replacing-the-list-monoid.md
```diff
@@ -0,0 +1,66 @@
...
+This is because, with `List`, appending a single element requires iterating over the entire data structure and therefore takes linear time.
+
+So `List` isn't all that great for this use case, so let's use `Vector` or `NonEmptyVector` instead, right?
+
+Well, `Vector` has its own problems and in this case it's unfortunately not all that faster than `List` at all. You can check [this blog post](http://www.lihaoyi.com/post/BenchmarkingScalaCollections.html#vectors-are-ok) by Li Haoyi for some deeper insight into `Vector`s issues.
+
+Because of this, it's now time to welcome a new data structure to Cats.
+Meet `Chain` and it's non-empty counterpart `NonEmptyChain`. 
+
+Available in the newest Cats 1.3.0 release, `Chain` is somewhat derived from what used to be `fs2.Catenable` and Erik Osheim's [Chain](https://github.com/non/chain ) library.
```
SystemFw: `somewhat derived` --> `evolved from`? A bit more enthusiastic :joy:

posts/2018-09-04-chain-replacing-the-list-monoid.md
```diff
@@ -0,0 +1,66 @@
...
+Similar to `List` it is also a very simple data structure, but unlike `List` it supports both constant time `append` and `prepend`.
```
SystemFw: > guarantees constant time for both `append` and `prepend`?

Always unsure on wheter to prefer `O(1)` or `constant`

posts/2018-09-04-chain-replacing-the-list-monoid.md
```diff
@@ -0,0 +1,66 @@
...
+This makes its `Monoid` instance super performant and a much better fit for usage with `Validated`,`Writer`, `Ior` or `Const`.
+
+To utilize this, we've added a bunch of shorthands in Cats 1.3 that previously used `NonEmptyList` to use `NonEmptyChain` instead. These include type aliases like `ValidatedNec` or `IorNec` as well as helper functions like `groupByNec` or `Validated.invalidNec`.
```
SystemFw: "We've changed a bunch .... to use"

posts/2018-09-04-chain-replacing-the-list-monoid.md
LukaJCB: I'll just add both :P

posts/2018-09-04-chain-replacing-the-list-monoid.md
LukaJCB: Hm, on the one hand you're right and the sentence is not well thought out, but we didn't change the shorthands in a breaking way, but just added new ones that mirror the old ones. I'll change the wording :)

posts/2018-09-04-chain-replacing-the-list-monoid.md
```diff
@@ -0,0 +1,66 @@
...
+[info] ChainBench.foldLeftLargeList       thrpt   20        135.954 ±       3.340  ops/s
+[info] ChainBench.foldLeftLargeVector     thrpt   20         61.613 ±       1.326  ops/s
+[info]
+[info] ChainBench.mapLargeChain           thrpt   20         59.379 ±       0.866  ops/s
+[info] ChainBench.mapLargeList            thrpt   20         66.729 ±       7.165  ops/s
+[info] ChainBench.mapLargeVector          thrpt   20         61.374 ±       2.004  ops/s
+```
+
+While not as dominant, `Chain` can still hold its ground fairly well. 
+It won't have the random access performance of something like `Vector`, but in almost all other cases, `Chain` seems to outperform it quite handily.
```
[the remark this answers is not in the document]
LukaJCB: How about "in a lot of other cases, `Chain` seems to perform on par or better."?

LukaJCB:
@kailuowang I was going to add some more technical documentation including how exactly it works under the hood :)

larsrh:
Finally, Typelevel has a blog chain. :laughing:

posts/2018-09-04-chain-replacing-the-list-monoid.md
```diff
@@ -0,0 +1,66 @@
...
+
+Traversing a data structure with something like `Writer[List[Log], A]` or `ValidatedNel[Error, A]` is  powerful and allows us to precisely specify what kind of iteration we want to do while remaining succint.
+However, in terms of efficiency it's a whole different story unfortunately.
+That is because both of these traversals make use of the `List` monoid (or the `NonEmptyList` semigroup), which by the nature of `List` is very inefficient.
+If you use `traverse` with a data structure with `n` elements and `Writer` or `Validated` as the `Applicative` type, you will end up with a runtime of `O(n^2)`.
+This is because, with `List`, appending a single element requires iterating over the entire data structure and therefore takes linear time.
+
+So `List` isn't all that great for this use case, so let's use `Vector` or `NonEmptyVector` instead, right?
+
+Well, `Vector` has its own problems and in this case it's unfortunately not that much faster than `List` at all. You can check [this blog post](http://www.lihaoyi.com/post/BenchmarkingScalaCollections.html#vectors-are-ok) by Li Haoyi for some deeper insight into `Vector`s issues.
```
tpolecat: Missing apostrophe: `Vector`'s

posts/2018-09-04-chain-replacing-the-list-monoid.md
```diff
@@ -0,0 +1,66 @@
...
+
+Because of this, it's now time to welcome a new data structure to Cats.
+Meet `Chain` and it's non-empty counterpart, `NonEmptyChain`.
```
tpolecat: its

PR #233, opened 2018-09-30
Add a blog post about variance and monad transformers

ceedubs:
Every once in a while [people](https://github.com/typelevel/cats/issues/556) [ask](https://github.com/typelevel/cats/issues/2310) [about](https://github.com/typelevel/cats/issues/2538) this.

I thought that it would be nice to provide a more detailed answer in a
single place so we can link to it from the Cats FAQ.

Note: I had to add the `-Ypartial-unification` flag for one of the code
samples to compile. It's quite possible that I should be doing this a
different way than I did.

### DIFF
--- a/posts/2018-09-29-monad-transformer-variance.md
+++ b/posts/2018-09-29-monad-transformer-variance.md
@@ -1,0 +1,161 @@
+---
+layout: post
+title: Monad Transformer Variance
+
+meta:
+  nav: blog
+  author: ceedubs
+  pygments: true
+
+tut:
+  scala: 2.12.7
+  binaryScala: "2.12"
+  dependencies:
+    - org.scala-lang:scala-library:2.12.7
+    - org.typelevel::cats-core:1.4.0
+    - io.circe::circe-core:0.9.3
+
+---
+
+A question that repeatedly pops up about [Cats](https://typelevel.org/cats/) is why monad transformer types like `OptionT` and `EitherT` aren't covariant like their `Option` and `Either` counterparts. This blog post aims to answer that question.
+
+# Covariance
+
+What does it mean to say that `Option` is covariant? It means that an `Option[B]` is allowed to be treated as an `Option[A]` if `B` is a subtype of `A`. For example:
+
+```tut:silent
+abstract class Err(val msg: String)
+final case class NotFound(id: Long) extends Err(s"Not found: $id")
+```
+
+```tut:book
+val optionNotFound: Option[NotFound] = Some(NotFound(42L))
+
+optionNotFound: Option[Err]
+```
+
+Great. If you want to treat your `Option[Notfound]` as an `Option[Err]`, you are free to.
+
+This is made possible because the `Option` type is declared as `sealed abstract class Option[+A]`, where the `+` in front of the `A` means that it is covariant in the `A` type parameter.
+
+What happens if we try to do the same with `OptionT`?
+
+```tut:silent
+import cats.Eval
+import cats.data.OptionT
+```
+
+```tut:book
+val optionTNotFound: OptionT[Eval, NotFound] = OptionT(Eval.now(optionNotFound))
+```
+
+```tut:book:fail
+optionTNotFound: OptionT[Eval, Err]
+```
+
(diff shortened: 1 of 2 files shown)


posts/2018-09-29-monad-transformer-variance.md
```diff
@@ -0,0 +1,161 @@
+---
+layout: post
+title: Monad Transformer Variance
```
larsrh: I suggest "Variance of Monad Transformers"

posts/2018-09-29-monad-transformer-variance.md
```diff
@@ -0,0 +1,161 @@
...
+  scala: 2.12.7
+  binaryScala: "2.12"
+  dependencies:
+    - org.scala-lang:scala-library:2.12.7
+    - org.typelevel::cats-core:1.4.0
+    - io.circe::circe-core:0.9.3
+
+---
+
+A question that repeatedly pops up about [Cats](https://typelevel.org/cats/) is why monad transformer types like `OptionT` and `EitherT` aren't covariant like their `Option` and `Either` counterparts. This blog post aims to answer that question.
```
larsrh: Can you link to these issues here?

posts/2018-09-29-monad-transformer-variance.md
```diff
@@ -0,0 +1,161 @@
...
+final case class NotFound(id: Long) extends Err(s"Not found: $id")
+```
+
+```tut:book
+val optionNotFound: Option[NotFound] = Some(NotFound(42L))
+
+optionNotFound: Option[Err]
+```
+
+Great. If you want to treat your `Option[Notfound]` as an `Option[Err]`, you are free to.
```
larsrh: `NotFound`

posts/2018-09-29-monad-transformer-variance.md
```diff
@@ -0,0 +1,161 @@
...
+```
+
+However, we can't do the same with our `CovariantOptionT`, because we can't even create an `OptionT` where the `F` type isn't covariant:
+
+```tut:book:fail
+def wrap[A](optionDecoder: Decoder[Option[A]]): CovariantOptionT[Decoder, A] =
+  CovariantOptionT[Decoder, A](optionDecoder)
+```
+
+In this particular case, `Decoder` _could_ be declared as covariant, but it's not. It would be unfortunate to lose the ability to use a monad transformer because a 3rd party library chose not to make a type covariant. And perhaps more importantly, sometimes you might want to use an `OptionT` with an `F` type that fundamentally isn't covariant in nature, such as `Monoid` (which is invariant) or `Order` (which is contravariant in nature and is declared as invariant in its type definition).
```
larsrh: Is this really a good argument without further context? At this point I think it makes sense to point out that `OptionT` also has use cases for contravariant type constructors, like a `contramap` function.

larsrh:
Thanks @ceedubs for writing that up!

ceedubs:
Thanks for the great review and handling the flag feature @larsrh! I'm hoping to get to your comments at some point today (US Pacific time).

posts/2018-09-29-monad-transformer-variance.md
ceedubs: Yeah, I had a whole section written up here that was using `Show` and then ran into https://github.com/typelevel/cats/issues/2545 😿 . I became hesitant to emphasize contravariant or invariant functors too much because in reality `OptionT` doesn't really support those well, and people who tried would run into issues. Also even assuming that those things existed, I was struggling to come up with examples using `OptionT` that didn't feel really contrived but didn't bring in too much other context. If you have have any good ideas, please give them. Also I may be able to just bring up the example that I have further down with `Eq`. I can look into rearranging a bit and providing better motivation.

larsrh: Fair point. In that case, I think a potentially weak argument is better than something that requires instances that don't exist yet :smile:

ceedubs:
@larsrh I've made an attempt to address your feedback. Please let me know what you think.

larsrh:
Thank you for your contribution!

PR #237, opened 2018-10-28
Add Semiring post

LukaJCB:
Wrote this up on my flight back from LambdaWorld, too tired right now to add any more description, but feedback welcome as always 😴

### DIFF
--- a/posts/2018-11-02-semirings.md
+++ b/posts/2018-11-02-semirings.md
@@ -23,13 +23,11 @@ tut:
 
 
 
-Ever wondered why sum types are called sum types?
+*Ever wondered why sum types are called sum types?
 Or maybe you've always wondered why the `<*>` operator uses exactly these symbols?
 And what do these things have to do with Semirings?
-Read this article and find out!
+Read this article and find out!*
 
-
-----
 We all know and use `Monoid`s and `Semigroup`s.
 They're super useful and come with properties that we can directly utilize to gain a higher level of abstractions at very little cost.
 Sometimes, however, certain types can have multiple `Monoid` or `Semigroup` instances.
@@ -65,13 +63,13 @@ import simulacrum._
 
 A `Semiring` is then just an `AdditiveMonoid` coupled with a `MultiplicativeMonoid` with the following extra laws:
 
-1. Additive commutativity. i.e. `x + y === y + x`
-2. Right Distributivity. i.e. `(x + y) * z === (x * z) + (y * z)`
-3. Right Distributivity. i.e. `x * (y + z) === (x * y) + (x * z)`
-4. Right absorption. i.e. `x * zero === zero`
-5. Left absorption. i.e. `zero * x === zero`
+1. Additive commutativity, i.e. `x + y === y + x`
+2. Right distributivity, i.e. `(x + y) * z === (x * z) + (y * z)`
+3. Left distributivity, i.e. `x * (y + z) === (x * y) + (x * z)`
+4. Right absorption, i.e. `x * zero === zero`
+5. Left absorption, i.e. `zero * x === zero`
 
-To define it as a typeclass, we simply extend from both additive and multiplicative monoid:
+To define it as a type class, we simply extend from both additive and multiplicative monoid:
 
 ```tut:silent
 @typeclass trait Semiring[A] extends MultiplicativeMonoid[A] with AdditiveMonoid[A]
@@ -84,15 +82,17 @@ One very interesting thing, I'd like to point out is the fact that we can form a
 What the hell is that?
 Well, bear with me for a while and I'll try to explain step by step.
 
+### Cardinality
+
 Okay, so let's start with what I mean by cardinality.
-Every type has a specific number of values it can possibly have, e.g. a `Boolean` has a cardinality of 2, because it has two possible values: `true` and `false`.
+Every type has a specific number of values it can possibly have, e.g. a `Boolean` has cardinality of 2, because it has two possible values: `true` and `false`.
 
 So `Boolean` has two, how many do other primitive types have?
 `Byte` has 2^8, `Short` has 2^16, `Int` has 2^32 and `Long` has 2^64.
 So far so good, that makes sense, what about something like `String`?
 `String` is an unbounded type and therefore theoretically has infinite number of different inhabitants (practically of course, we don't have infinite memory, so the actual number may vary depending on your system).
 
-For what other types can we determine their number of possible inhabitants? 
+For what other types can we determine their cardinality?
 Well a couple of easy ones are `Unit`, which has exactly one value it can take and also `Nothing`, which has 0 possible values.
 
 That's neat, maybe we can encode this in actual code.
@@ -114,31 +114,31 @@ Awesome!
 Now let's try to define some instances for this type class:
 
 ```tut:silent
-implicit def booleanNumberOfInhabitants = new Cardinality[Boolean] {
+implicit def booleanCardinality = new Cardinality[Boolean] {
   def cardinality: BigInt = BigInt(2)
 }
 
-implicit def longNumberOfInhabitants = new Cardinality[Long] {
+implicit def longCardinality = new Cardinality[Long] {
   def cardinality: BigInt = BigInt(2).pow(64)
 }
 
-implicit def intNumberOfInhabitants = new Cardinality[Int] {
+implicit def intCardinality = new Cardinality[Int] {
   def cardinality: BigInt = BigInt(2).pow(32)
 }
 
-implicit def shortNumberOfInhabitants = new Cardinality[Short] {
+implicit def shortCardinality = new Cardinality[Short] {
   def cardinality: BigInt = BigInt(2).pow(16)
 }
 
-implicit def byteNumberOfInhabitants = new Cardinality[Byte] {
+implicit def byteCardinality = new Cardinality[Byte] {
   def cardinality: BigInt = BigInt(2).pow(8)
 }
 
-implicit def unitNumberOfInhabitants = new Cardinality[Unit] {
+implicit def unitCardinality = new Cardinality[Unit] {
   def cardinality: BigInt = 1
 }
 
-implicit def nothingNumberOfInhabitants = new Cardinality[Nothing] {
+implicit def nothingCardinality = new Cardinality[Nothing] {
   def cardinality: BigInt = 0
 }
 ```
@@ -170,7 +170,7 @@ If you try this with other examples, you'll see that it's exactly true, awesome!
 Let's encode that fact in a type class instance:
 
 ```tut:silent
-implicit def tupleNumberOfInhabitants[A: Cardinality, B: Cardinality] =
+implicit def tupleCardinality[A: Cardinality, B: Cardinality] =
   new Cardinality[(A, B)] {
     def cardinality: BigInt = Cardinality[A].cardinality * Cardinality[B].cardinality
   }
@@ -183,7 +183,7 @@ So `Either[Boolean, Byte]` should have `2 + 256 = 258` number of inhabitants. Co
 Let's also code that up and try and confirm what we learned in the REPL:
 
 ```tut:silent
-implicit def eitherNumberOfInhabitants[A: Cardinality, B: Cardinality] =
+implicit def eitherCardinality[A: Cardinality, B: Cardinality] =
   new Cardinality[Either[A, B]] {
     def cardinality: BigInt = Cardinality[A].cardinality + Cardinality[B].cardinality
   }
@@ -202,7 +202,7 @@ That makes a lot of sense given their names!
 
 
 So what about that homomorphism we talked about earlier? 
-Well, a homomorphism is a structure-preserving mapping function between two algebraic structures of the same type (in this case a semiring).
+Well, a homomorphism is a structure-preserving mapping function between two algebraic structures of the same sort (in this case a semiring).
 
 This means that for any two values `x` and `y` and the homomorphism `f`, we get 
 1. `f(x * y) === f(x) * f(y)`
@@ -425,11 +425,12 @@ implicit def constSemiringal[A: Semiring] = new Semiringal[Const[A, ?]] {
 Rings and Semirings are very interesting algebraic structures and even if we didn't know about them we've probably been using them for quite some time.
 This blog post aimed to show how `Applicative` and `MonoidK` relate to `Monoid` and how algebraic data types form a semiring and how these algebraic structures are pervasive throughout Scala and other functional programming languages.
 For me personally, realizing how all of this ties together and form some really satisfying symmetry was really mind blowing and I hope this blog post can give some good insight on recognizing these interesting similarities throughout Cats and other libraries based on different mathematical abstractions.
+For further material on this topic, you can check out [this talk](https://www.youtube.com/watch?v=YScIPA8RbVE).
 
 
 
 #### Addendum:
 
-This article glossed over commutativity in the typeclass encodings.
+This article glossed over commutativity in the type class encodings.
 Commutativity is very important law for semrings and the code should show that.
 However, since this post already contained a lot of different type class definitions, adding extra commutative type class definitions that do nothing but add laws felt like it would distract from what is trying to be taught.
--- a/posts/2018-11-02-semirings.md
+++ b/posts/2018-11-02-semirings.md
@@ -18,7 +18,6 @@ tut:
   dependencies:
     - com.github.mpilquist::simulacrum:0.14.0
   plugins:
-    - org.spire-math::kind-projector:0.9.8
     - org.scalamacros:paradise_2.12.7:2.1.0
 ---
 
@@ -120,7 +119,7 @@ implicit def booleanNumberOfInhabitants = new PossibleInhabitants[Boolean] {
 }
 
 implicit def longNumberOfInhabitants = new PossibleInhabitants[Long] {
-  override def number: BigInt = BigInt(2).pow(64)
+  def number: BigInt = BigInt(2).pow(64)
 }
 
 implicit def intNumberOfInhabitants = new PossibleInhabitants[Int] {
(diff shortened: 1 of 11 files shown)


larsrh:
I'm seeing some really weird tut errors. Let's see if this patch fixes it.

larsrh:
Did this compile locally, @LukaJCB? There may be a weird problem with kind-projector here.

```
[tut] *** Error reported at /home/travis/build/typelevel/typelevel.github.com/posts/2018-11-02-semirings.md:421
<console>:36: error: not found: type ?
       implicit def constSemiringal[A: Semiring] = new Semiringal[Const[A, ?]] {
                                                                           ^
<console>:36: error: Const[A,<error>] takes no type parameters, expected: one
       implicit def constSemiringal[A: Semiring] = new Semiringal[Const[A, ?]] {
                                                                  ^
```

LukaJCB:
That seems very odd, no clue what might cause this. And no, I haven't tried locally yet

posts/2018-11-02-semirings.md
```diff
@@ -0,0 +1,436 @@
...
+
+Indeed, and we can show that quite easily.
+
+`MonoidK` adds an `empty` function with the following definition:
+
+```scala
+def empty[A]: F[A]
+```
+
+This function uses an existential quantifier for `A`, which means that it works for any `A`, which then means that it cannot actually include any particular `A` and is therefore equivalent to `F[Nothing]` which is what we have for `AdditiveMonoidal`.
```
smarter: existential -> universal

larsrh:
Well, it's the last tut block, so in principle, it could just be a scala block instead.

posts/2018-11-02-semirings.md
```diff
@@ -0,0 +1,435 @@
...
+Optimally it'd be something like `[+]`, but that's not a valid identifier so let's try `<+>` instead!
+
+```scala
+def <+>[A, B](fa: F[A], fb: F[B]): F[Either[A, B]]
+```
+
+Oh! The `<+>` function already exists in cats as an alias for `combineK` which can be found on `SemigroupK`, but it's sort of different, it takes two `F[A]`s and returns an `F[A]`, not quite what we have here.
+
+Or is it?
+These two functions are actually the same, and we can define them in terms of one another as long as we have an invariant functor:
```
iravid: Did you intend to show the other direction too?

posts/2018-11-02-semirings.md
```diff
@@ -0,0 +1,435 @@
...
+@typeclass trait MultiplicativeMonoid[A] extends MultiplicativeSemigroup[A] {
+  def one: A
+}
+```
+
+A `Semiring` is then just an `AdditiveMonoid` coupled with a `MultiplicativeMonoid` with the following extra laws:
+
+1. Additive commutativity. i.e. `x + y === y + x`
+2. Right Distributivity. i.e. `(x + y) * z === (x * z) + (y * z)`
+3. Right Distributivity. i.e. `x * (y + z) === (x * y) + (x * z)`
```
rkrzewski:
```suggestion
3. Left Distributivity. i.e. `x * (y + z) === (x * y) + (x * z)`
```

posts/2018-11-02-semirings.md
LukaJCB: No, I think just showing one should be fine, blog post is already plenty long as is :)

posts/2018-11-02-semirings.md
LukaJCB: Thanks!

a-reisberg:
Great post! I have some comments. Sorry in advance for the use of non-standard notations (just to make thing shorter). And sorry for such a long comment. I thought it was a quick one...

Your use of `Invariant` throws me off a bit.  It seems like just functor suffice (maybe I've missed something):
- Given a functor `F` with `F[A] x F[B] => F[A + B]` (combine) for all types `A, B`, we get sum like you do: `F[A] x F[A] => F[A+A] => F[A]`, where the first arrow is combine and the second one is merge and functoriality of `F`.
- Given a functor `F` with `F[A] x F[A] => F[A]` (sum), we get back combine like so: `F[A] x F[B] => F[A + B] x F[A + B] => F[A + B]` where the first arrow is given by inserting to left (resp. right) for A (resp. B) into `A + B` as well as the functoriality of `F`, and the second arrow is just sum. This also answers @iravid 's question.

In general, I think that the use of invariant functors in the post is a bit strange. And the choice of multiplicative unit is also a bit hard to motivate, at least, from the categorical standpoint, which might be why invariant functors crop up. I would propose the following (categorical motivations to follow)
- For multiplicative monoidal, the unit is still `A => F[A]` as in Applicatives.
- For additive monoidal, the unit is the same as what you already have.
So essentially, all I'm saying is that cats already have these in the exact same form, but just with a different names. :)

Here are the categorical motivation. Essentially we are talking about lax monoidal functors from `Scal` to `Scal`. See https://ncatlab.org/nlab/show/monoidal+functor for quick definitions of lax monoidal. For multiplicative monoidal, it's `(Scal, +)` to `(Scal, x)` whereas for the additive one, it's `(Scal, +)` to `(Scal, x)`. The lax monoidal aspect (modulo units) gets translated to 
- `F[A] x F[B] => F[A x B]` for the multiplicative one
- `F[A] x F[B] => F[A + B]` for the additive one

and the units are:

- `(() => A) => (() => F[A])` for the multiplicative one. Here, I use `()` to denote the type with only 1 inhabitant. So, `() => A` corresponds to a choice of an element `a` in `A`, and `() => F[A]` an element `pure(a)` in `F(A)`. So this is exactly `A => F[A]`. Essentially, in nLab's terms, we are requiring that `F` sends `()` to `()`. 
- Following the above, for the additive one, we want `F[Nothing] = ()`. Applying the same trick, taking into account of the fact that `Nothing` is initial in `Scal` (i.e. there's exactly one map `Nothing => A` for all types `A`), we get `() => F[A]` for all types `A`, i.e. a choice of an element of `F[A]` for each type `A`. This is why `emptyK` has such a signature.

_Side note:_ the statements at the beginning amounts to saying that lax monoidal preserves monoid objects. Indeed, for an additive monoidal `F` (i.e. lax monoidal wrt to `+` and `x` on `Scal`), `F[A]` is always a monoid because `A` is always a monoid wrt. to `+` (via `merge`).

You probably already know that lax monoidal functors are indeed monoids in the category of functors, where in our case, the category of functors `Fun(Scal, Scal)` can be equipped with a Day convolution monoidal structure for each choice of a pair of monoidal structures on the source and target `Scal`. So really, we've been talking about monoids all along, just within different categories.

LukaJCB:
Thanks a ton for the feedback, it's great to have a perspective from a mathematician on this and I really appreciate the detail you put into your post :)

> Your use of Invariant throws me off a bit. It seems like just functor suffice (maybe I've missed something):

Right functor definitely suffices, but an invariant functor is more general.
We already have invariant and contravariant definitions of `Applicative` and `Apply`, so to me it stands to reason that there could be Additive Monoidal and Semigroupal Invariant and Contravariant Functors. The Invariant ones being generalizations of both the co and contravariant ones. 
Maybe that's another blog post :P
Though Haskell for example has [Decidable](http://hackage.haskell.org/package/contravariant-1.5/docs/Data-Functor-Contravariant-Divisible.html#t:Decidable) which is basically a Contravariant Semiringal functor. 

> For multiplicative monoidal, the unit is still A => F[A] as in Applicatives.

I use the different form of `F[Unit]` to make the parallels more obvious, we learned in the section before that `Unit` behaves like the multiplicative identity, so I think that makes it more clear IMO.

> So essentially, all I'm saying is that cats already have these in the exact same form, but just with a different names. :)

Hey, that's what I'm saying too! :)

> For multiplicative monoidal, it's (Scal, +) to (Scal, x) whereas for the additive one, it's (Scal, +) to (Scal, x). 

I don't quite understand what you mean with this, aren't they the same? 
`(Scal, +) => (Scal, x)`?

> You probably already know that lax monoidal functors are indeed monoids in the category of functors, where in our case, the category of functors Fun(Scal, Scal) can be equipped with a Day convolution monoidal structure for each choice of a pair of monoidal structures on the source and target Scal. So really, we've been talking about monoids all along, just within different categories.

I'm not quite sure what exactly Day convolution is, but I think I understand what you're saying :)

Hope that answers your questions for now :+1:

a-reisberg:
> Right functor definitely suffices, but an invariant functor is more general.
We already have invariant and contravariant definitions of Applicative and Apply, so to me it stands to reason that there could be Additive Monoidal and Semigroupal Invariant and Contravariant Functors. The Invariant ones being generalizations of both the co and contravariant ones.
Maybe that's another blog post :P

Right! I never thought of it that way since usually (i.e. outside of programming for me ;) ), invariants functors are themselves just functors, where we replace `Scal` by something else slightly different. So all the reasonings about lax monoidal, monoids etc. go as for usual functors, without mentioning invariant. Unfortunately, this thinking doesn't really carry over verbatim since we don't have categories in Scala.

> I use the different form of F[Unit] to make the parallels more obvious, we learned in the section before that Unit behaves like the multiplicative identity, so I think that makes it more clear IMO. Hey, that's what I'm saying too! :)

Yup! It was just me not reading the first part. :)

> I don't quite understand what you mean with this, aren't they the same?
(Scal, +) => (Scal, x)?

Ooops, that was a typo. The sentence should be

> For multiplicative monoidal, it's (Scal, x) to (Scal, x) whereas for the additive one, it's (Scal, +) to (Scal, x). 

And again, we unfortunately don't have categories in Scala, otherwise, we can define what a lax monoidal functor is, and then define different monoidal structures on `Scal` to automatically get both the multiplicative and additive versions :).

For Day convolutions, you can take a look at 
https://bartoszmilewski.com/2017/02/06/applicative-functors/
where it's defined via ends/coends. Personally though, I prefer defining it as a left Kan extension, since that's closer to usual convolutions (say, in analysis).

LukaJCB:
Seems like we're pretty much in agreement then :) Cool, thanks again! Also thanks for that blog post :)

posts/2018-11-02-semirings.md
```diff
@@ -0,0 +1,435 @@
...
+
+
+
+Ever wondered why sum types are called sum types?
+Or maybe you've always wondered why the `<*>` operator uses exactly these symbols?
+And what do these things have to do with Semirings?
+Read this article and find out!
+
+
+----
```
larsrh: Please don't use these horizontal rulers. Instead, emphasize the introductory text (italics, for example).

posts/2018-11-02-semirings.md
```diff
@@ -0,0 +1,435 @@
...
+They're super useful and come with properties that we can directly utilize to gain a higher level of abstractions at very little cost.
+Sometimes, however, certain types can have multiple `Monoid` or `Semigroup` instances.
+An easy example are the various numeric types where both multiplication and addition form two completely lawful monoid instances.
+
+In abstract algebra there is a an algebraic class for types with two `Monoid` instances that interact in a certain way.
+These are called `Semiring`s (sometimes also `Rig`) and they are defined as two `Monoid`s with some special laws that define the interactions between them.
+Because they are often used to describe numeric data types we usually classify them as *Additive* and *Multiplicative*. 
+Just like with numeric types the laws of `Semiring` state that multiplication has to distribute over addition and multiplying a value with the additive identity (i.e. zero) absorbs the value and becomes zero.
+
+There are different ways to encode this as type classes and different libraries handle this differently, but let's look at how the `typelevel/algebra` project handles this.
```
larsrh:
```suggestion
There are different ways to encode this as type classes and different libraries handle this differently, but let's look at how the [algebra](https://typelevel.org/algebra/) project handles this.
```

posts/2018-11-02-semirings.md
```diff
@@ -0,0 +1,435 @@
...
+}
+```
+
+A `Semiring` is then just an `AdditiveMonoid` coupled with a `MultiplicativeMonoid` with the following extra laws:
+
+1. Additive commutativity. i.e. `x + y === y + x`
+2. Right Distributivity. i.e. `(x + y) * z === (x * z) + (y * z)`
+3. Left Distributivity. i.e. `x * (y + z) === (x * y) + (x * z)`
+4. Right absorption. i.e. `x * zero === zero`
+5. Left absorption. i.e. `zero * x === zero`
```
larsrh: lower case "Distributivity" to be consistent; use comma before i.e.

posts/2018-11-02-semirings.md
```diff
@@ -0,0 +1,435 @@
...
+To define it as a typeclass, we simply extend from both additive and multiplicative monoid:
+
+```tut:silent
+@typeclass trait Semiring[A] extends MultiplicativeMonoid[A] with AdditiveMonoid[A]
+```
+
+Now we have a `Semiring` class, that we can use with the various numeric types like `Int`, `Long`, `BigDecimal` etc, but what else is a `Semiring` and why dedicate a whole blog post to it?
+
+It turns out a lot of interesting things can be `Semiring`s, including `Boolean`s, `Set`s and [animations](https://bkase.github.io/slides/algebra-driven-design/#/).
+One very interesting thing, I'd like to point out is the fact that we can form a Semiring homomorphism from types to their number of possible inhabitants.
```
larsrh:
```suggestion
One very interesting thing I'd like to point out is that we can form a `Semiring` homomorphism from types to their number of possible inhabitants.
```

posts/2018-11-02-semirings.md
```diff
@@ -0,0 +1,435 @@
...
+What the hell is that?
+Well, bear with me for a while and I'll try to explain step by step.
+
+Okay, so let's start with what I mean by possible inhabitants.
```
larsrh: Maybe add some section headings?

posts/2018-11-02-semirings.md
```diff
@@ -0,0 +1,435 @@
...
+Every type has a specific number of values it can possibly have, e.g. a `Boolean` has 2 possible inhabitants, because it has two possible values: `true` and `false`.
+
+So `Boolean` has two, how many do other primitive types have?
+`Byte` has 2^8, `Short` has 2^16, `Int` has 2^32 and `Long` has 2^64.
+So far so good, that makes sense, what about something like `String`?
+`String` is an unbounded type and therefore theoretically has infinite number of different inhabitants (practically of course, we don't have infinite memory, so the actual number may vary depending on your system).
+
+For what other types can we determine their number of possible inhabitants? 
+Well a couple of easy ones are `Unit`, which has exactly one value it can take and also `Nothing`, which has 0 possible values.
```
larsrh: I think `Nothing` warrants some more explanation here. Readers may not be aware of how it works. Also, `null` is generally ignored if you want to play by the rules.

posts/2018-11-02-semirings.md
```diff
@@ -0,0 +1,435 @@
...
+
+That's neat, maybe we can encode this in actual code.
+We could create a type class that should be able to give us the number of inhabitants for any type we give it:
```
larsrh: I think you're inconsistent with "type class" vs "typeclass".

posts/2018-11-02-semirings.md
```diff
@@ -0,0 +1,435 @@
...
+
+```tut:silent
+trait PossibleInhabitants[A] {
+  def number: BigInt
+}
+
+object PossibleInhabitants {
+  def of[A: PossibleInhabitants]: BigInt = apply[A].number
```
larsrh: Is this method really needed? `PossibleInhabitants[Int].number` is not much longer than `PossibleInhabitants.of[Int]`.

posts/2018-11-02-semirings.md
```diff
@@ -0,0 +1,435 @@
...
+
+PossibleInhabitants.of[Either[Int, (Boolean, Unit)]]
+```
+
+So using sum types seem to add the number of inhabitants whereas product types seem to multiply the number of inhabitants.
+That makes a lot of sense given their names!
+
+
+So what about that homomorphism we talked about earlier? 
+Well, a homomorphism is a structure-preserving mapping function between two algebraic structures of the same type (in this case a semiring).
```
larsrh: Not sure about "same type". Maybe "same sort"? "same gestalt"? "same shape"?

posts/2018-11-02-semirings.md
```diff
@@ -0,0 +1,435 @@
...
+It is! Since `Nothing` doesn't have any values an `Either[Nothing, A]` can only be a `Right` and therefore only an `A`, so these are in fact equivalent types.
+
+We also have to check for the absorption law that says that any value mutliplied with the additive identity `zero` should be equivalent to `zero`.
+Since `Nothing` is our `zero` a product type like `(Int, Nothing)` should be equivalent to `Nothing`.
+This also holds, given the fact that we can't construct a `Nothing` so we can never construct a tuple that expects a value of type `Nothing` either.
+
+Let's see if this translates to the number of possible inhabitants as well:
+
+Additive Identity:
+```scala
```
larsrh: tut?

posts/2018-11-02-semirings.md
```diff
@@ -0,0 +1,435 @@
...
+
+## Conclusion
+
+Rings and Semirings are very interesting algebraic structures and even if we didn't know about them we've probably been using them for quite some time.
+This blog post aimed to show how `Applicative` and `MonoidK` relate to `Monoid` and how algebraic data types form a semiring and how these algebraic structures are pervasive throughout Scala and other functional programming languages.
+For me personally, realizing how all of this ties together and form some really satisfying symmetry was really mind blowing and I hope this blog post can give some good insight on recognizing these interesting similarities throughout Cats and other libraries based on different mathematical abstractions.
+
+
+
+#### Addendum:
```
larsrh:
```suggestion
## Addendum
```

I like consistency :sweat_smile:

larsrh:
Another thing you could link to is this talk: https://www.youtube.com/watch?v=YScIPA8RbVE

larsrh:
Thanks for bearing with me :sweat_smile:

LukaJCB:
I'm gonna go ahead and merge this one :)

posts/2018-11-02-semirings.md
```diff
@@ -0,0 +1,438 @@
...
+---
+
+
+
+*Ever wondered why sum types are called sum types?
+Or maybe you've always wondered why the `<*>` operator uses exactly these symbols?
+And what do these things have to do with Semirings?
+Read this article and find out!*
+
+We all know and use `Monoid`s and `Semigroup`s.
```
mpilquist: Minor comment: for folks that don't know or use these, this sentence may drive them off. I suggest rewording with a link to some background info if the reader isn't comfortable with these concepts.

LukaJCB: Yeah, agreed :+1:

PR #241, opened 2018-11-27
[Blog Post] - Error handling in http4s with classy optics - Part 2

gvolpe:


posts/2018-11-28-http4s-error-handling-mtl-2.md
```diff
@@ -0,0 +1,456 @@
+---
+layout: post
+title: Error handling in Http4s with classy optics – Part 2
```
larsrh: Please use a title that's consistent with Part 1 ("Http4s error handling with Cats Meow MTL"). Also, it would be great to have a link from Part 1 to here.

posts/2018-11-28-http4s-error-handling-mtl-2.md
```diff
@@ -0,0 +1,456 @@
...
+      }
+  }
+
+  def routes(implicit H: HttpErrorHandler[F, UserError]): HttpRoutes[F] =
+    H.handle(httpRoutes)
+
+}
+```
+
+Notice that in contrary to the example shown in the previous blog post there is now a relationship between `UserAlg` and `HttpErrorHandler`: the error type is the same. However, this is not enforced by the compiler. ***Can we be more strict about it?***
```
larsrh: "contrary to" or "in contrast to"

larsrh: Don't combine bold face and italics.

posts/2018-11-28-http4s-error-handling-mtl-2.md
```diff
@@ -0,0 +1,456 @@
...
+    - io.circe::circe-generic:0.10.0
+    - co.fs2::fs2-core:1.0.0
+    - com.olegpy::meow-mtl:0.2.0
+    - com.chuusai::shapeless:2.3.3
+
+---
+
+This is a continuation of my [previous blog post](https://typelevel.org/blog/2018/08/25/http4s-error-handling-mtl.html). Make sure you have read that one before continuing here.
+
+I recently gave a 20 minutes talk on `classy optics` at the unconference of [Scale by the Bay](http://scale.bythebay.io/) where I also talked about this error handling technique and on my way back home I was still thinking of different ways of doing this. So, after some exploratory work, I came up with a few different alternatives.
```
larsrh: The first sentence in this paragraph is a bit long.

posts/2018-11-28-http4s-error-handling-mtl-2.md
```diff
@@ -0,0 +1,456 @@
...
+
+Neat! Right? If we try to change the error type of `UserAlg` it wouldn't compile!
+
+### More than one algebra per Http Route
+
+In most of my programs I tend to specify an `HttpRoute` per algebra. But what if we wanted to just define a single `HttpRoute` that uses multiple algebras? There are a couple of options.
+
+Let's first define a new ADT of errors and a new algebra to illustrate the problem:
+
+**Catalog Error**
```
larsrh: You can use four hashes for another indentation level.

posts/2018-11-28-http4s-error-handling-mtl-2.md
```diff
@@ -0,0 +1,456 @@
...
+             H2: HttpErrorHandler[F, CatalogError]
+  ): HttpRoutes[F] =
+    H2.handle(H1.handle(httpRoutes))
+
+}
+```
+
+It works! But it's not as elegant as we would like it to be and if we add more algebras this would quicky get out of control.
+
+***Can we generalize this pattern?***
```
larsrh: bold + italics

posts/2018-11-28-http4s-error-handling-mtl-2.md
```diff
@@ -0,0 +1,456 @@
...
+
+We can define our error type as a coproduct of different errors, in our case `UserError` and `CatalogError`. For example:
+
+```scala
+import shapeless._
+
+def routes[F[_]](implicit H: HttpErrorHandler[F, UserError :+: CatalogError :+: CNil]) = ???
+```
+
+However, this doesn't compile because the error type is no longer a subtype of `Throwable`. It is now a `Coproduct` :/. But...
```
larsrh: Neither ASCII emojis nor ellipses are particularly good style. Try to rephrase it slightly, e.g. join with the next paragraph.

posts/2018-11-28-http4s-error-handling-mtl-2.md
```diff
@@ -0,0 +1,456 @@
...
+
+### Conclusion
+
+The last approach is probably too much but we have demonstrated that it's possible to push the boundaries to make our application very type-safe. However, we also need to consider the trade-offs of writing more boilerplate.
+
+Personally, I settle for the previous approach where the error type of the algebra matches the error type of the `HttpErrorHandler` even if it requires a bit more of discipline. The choice is yours! Just make sure you understand the trade-offs of every mechanism.
+
+I hope you have enjoyed this post and please do let me know if you have other ideas to keep broadening my understanding!
+
+Thanks for reading :)
```
larsrh: Remove this sentence. Readers should be thanking you for writing the post ☺

posts/2018-11-28-http4s-error-handling-mtl-2.md
gvolpe: Do you think it's okay to update the title of the first part? I like this one more :smile: 

Do you mean adding a link to the title? How can I do that?

larsrh: > Do you think it's okay to update the title of the first part? I like this one more

No problem, you can add the change to this PR.

> Do you mean adding a link to the title? How can I do that?

I was thinking in the body text of the post. In this post, you have a back link somewhere in the beginning. You can do the same for the old post, but in the other direction.

gvolpe: Got it, will do soon! :+1:

gvolpe:
I believe I addressed all your comments @larsrh . 

I'd like to wait for a review from @rossabaker before it goes public in case that's possible :)

posts/2018-11-28-http4s-error-handling-mtl-2.md
```diff
@@ -0,0 +1,459 @@
...
+
+#### Data types with error handling capabilities
+
+As of now I still prefer the latter for a number of reasons. Here's a comparison table of all the different approaches I can think of:
+
+|            | EitherT[IO, E, ?] | IO[E, A] | IO[A] | IO[A] + Classy Optics |
+|------------|-------------------|----------|-------|-----------------------|
+| Concise code style | ✘ | ✔ | ✔ | ✔ |
+| Polymorphic code | ✔ | ✘ | ✔ | ✔ |
+| Performance overhead | **2x** | ✔ | ✔ | ✔ |
```
rossabaker: What happens in the `IO` almost always dominates the overhead of `EitherT`.  The benchmarks that make monad transformers look bad tend to be a worst case.

posts/2018-11-28-http4s-error-handling-mtl-2.md
```diff
@@ -0,0 +1,459 @@
...
+case class UserUpdateAge(age: Int)
+
+abstract class UserAlg[F[_]: ErrorChannel[?[_], E], E <: Throwable] {
+  def find(username: String): F[Option[User]]
+  def save(user: User): F[Unit]
+  def updateAge(username: String, age: Int): F[Unit]
+}
+```
+
+And here's the ADT of the possible errors that may arise (notice the `extends Exception` part):
```
rossabaker: We're assuming the entire audience knows that an `Exception` is a `Throwable`?

posts/2018-11-28-http4s-error-handling-mtl-2.md
```diff
@@ -0,0 +1,459 @@
...
+- `IO[E, A]` can either fail with an `E` or by throwing an exception when `A` was expected. This is called "unrecoverable" errors.
+
+And this is not a bad thing at all! It is like this by design and we can have the same property when writing polymorphic code using `cats-effect` while still keeping it simple. Here's one way:
+
+### Error Channel
+
+In the previous blog post we defined the algebras as a single trait. In this case we are going to try a different encoding but first we need to introduce an `ErrorChannel[F, E]` typeclass where the error type is a subtype of `Throwable` to be compatible with the error type of the `cats-effect` typeclasses:
+
+```tut:book:silent
+trait ErrorChannel[F[_], E <: Throwable] {
```
rossabaker: Why can't we just use `ApplicativeError` instead of inventing a new typeclass isomorphic to `ApplicativeError`?  Are we trying to avoid ambiguity with `ApplicativeError[F, Throwable]` from `F`'s other error channel?

posts/2018-11-28-http4s-error-handling-mtl-2.md
gvolpe: Would `~2x` be better? It's more an estimation than a concrete result given the extra depth of `flatMap` calls involved when using Monad Transformers.

posts/2018-11-28-http4s-error-handling-mtl-2.md
gvolpe: Two reasons why:

1. We can't have `ApplicativeError[F, UserError]` when we already have `Sync[F]` in scope that brings in `ApplicativeError[F, Throwable]`.
2. If we were to use `Sync[F].raiseError` we lose the type-safety we were trying to gain. The idea is that if somewhere in the call-chain there's a failure with `Throwable` in `F` that should be considered unrecoverable. Whereas if you use the `ErrorChannel` to raise errors you can be sure to handle your custom errors defined as an ADT. And you cannot do `ErrorChannel[F, UserError].raise(new Exception("boom"))` either as that won't compile.

posts/2018-11-28-http4s-error-handling-mtl-2.md
gvolpe: Added a comment below clarifying this.

gvolpe:
Hey @rossabaker any thoughts on this? I believe you are quite busy these days :smile_cat:

PR #242, opened 2018-12-06
Updating "error handling" blog post 

gvolpe:
I've got some feedback from John and after having a chat I'm updating the blog post accordingly.

We still have a pending discussion on "polymorphic code" but I think the only thing mentioned about it in this update is correct, however I would appreciate your review @jdegoes , thanks!

### DIFF
--- a/posts/2018-11-28-http4s-error-handling-mtl-2.md
+++ b/posts/2018-11-28-http4s-error-handling-mtl-2.md
@@ -47,7 +47,7 @@ The `IO[E, A]` model is naturally a better approach but I found out polymorphic 
 
 #### Errors vs Failures
 
-What I like about the `IO[E, A]` model is that we can distinguish between "business errors" and "unexpected failures" such as a database connection failure. Eg: when working on a REST API, most of the time we only care about mapping a few business errors to the appropriated http responses. The unexpected failures should be handled by someone else. In this case `http4s` will convert any failure into a response with code 500 (internal server error).
+What I like about the `IO[E, A]` model is that we can distinguish between "business errors" and "unexpected failures" such as a database connection failure. Eg: when working on a REST API, most of the time we only care about mapping a few business errors into the appropriated http responses. The unexpected failures should be handled by someone else. In this case `http4s` will convert any failure into a response with code 500 (internal server error).
 
 And this is exactly what we want to achieve here. Writing polymorphic code using `cats-effect` while trying to keep it as simple as possible. Here's an encoding I would like to explore further:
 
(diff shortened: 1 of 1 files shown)


posts/2018-11-28-http4s-error-handling-mtl-2.md
```diff
@@ -37,25 +37,19 @@ I recently gave a 20 minutes talk on `classy optics` at the unconference of [Sca
...
-|            | EitherT[IO, E, ?] | IO[E, A] | IO[A] | IO[A] + Classy Optics |
-|------------|-------------------|----------|-------|-----------------------|
-| Concise code style | ✘ | ✔ | ✔ | ✔ |
-| Polymorphic code | ✔ | ✘ | ✔ | ✔ |
-| Performance overhead | **~2x** | ✔ | ✔ | ✔ |
-| Double error channel | ✔ | ✔ | ✘ | ✘ |
+#### Errors vs Failures
 
-With "Double error channel" I mean that we can signal failure in two different ways:
+What I like about the `IO[E, A]` model is that we can distinguish between "business errors" and "unexpected failures" such as a database connection failure. Eg: when working on a REST API, most of the time we only care about mapping a few business errors into the appropriated http responses. The unexpected failures should be handled by someone else. In this case `http4s` will convert any failure into a response with code 500 (internal server error).
```
jdegoes: Much better! 👍 

I'd still quibble with this paragraph a little because `Either[E, A]` could be considered capable of distinguishing between "business errors" and "unexpected failures"—because you can have an unexpected failure anywhere you call a higher order function and pass a partial function.

The main differences is that when an `Either[E, A]` goes wrong, it will kill the thread running it (assuming you're not try / catching); whereas, when an `IO[E, A]` goes wrong, it will only kill the fiber running it. Both of course can be observed (in a sane and principled fashion) from the outside, but fibers are first-class functional constructs with typed return values, so it's trivial to supervise while still staying within the pure functional subset of Scala.

posts/2018-11-28-http4s-error-handling-mtl-2.md
```diff
@@ -37,25 +37,19 @@ I recently gave a 20 minutes talk on `classy optics` at the unconference of [Sca
...
+When working with `EitherT[F, E, A]` or a bifunctor `IO[E, A]` we have a clear error type whereas by just relying on a single `F[A]` with a `MonadError[F, Throwable]` instance we lose this property. There are a few issues with the first though:
 
-#### Data types with error handling capabilities
+- It has a "double error channel", meaning that can report errors via `Left` or via its effect type `IO`.
+- As any other Monad Transformer in Scala, introduces a performance overhead due to the extra `flatMap` calls and extra allocations.
+- Code becomes cumbersome as we need to lift effects and pure `Either` values into the transformer stack.
 
-As of now I still prefer the latter for a number of reasons. Here's a comparison table of all the different approaches I can think of:
+The `IO[E, A]` model is naturally a better approach but I found out polymorphic code is more cumbersome than working with `F[A]`.
```
jdegoes: It is more cumbersome with final tagless right now, although the main reason is that cats-effect doesn't have bifunctor type classes... yet. 🤞

posts/2018-11-28-http4s-error-handling-mtl-2.md
gvolpe: Thanks for your review @jdegoes ! I feel like adding all these details are out of the scope of this blog post so I just added a link to `zio`'s error model for further reading.

posts/2018-11-28-http4s-error-handling-mtl-2.md
gvolpe: Linked ;)

PR #244, opened 2019-01-24
Remove circe from Typelevel projects

travisbrown:
I don't really see any value in having circe listed as a Typelevel project. The removal here doesn't change anything about the way circe is maintained, or anything at all, really, apart from the fact that "circe is a Typelevel project" will no longer be true.

r? @larsrh @milessabin @non

### DIFF
--- a/_projects/circe.md
+++ b/_projects/circe.md
@@ -1,7 +1,0 @@
----
-layout: post
-title: "circe"
-category: "Principled JSON processing"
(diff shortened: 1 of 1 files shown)


travisbrown:
Putting on hold at @milessabin's request.

travisbrown:
Please merge. I'm preparing a blog post explaining the reasons for this decision (and the fact that it doesn't have any immediate practical implications), but I think the removal here should happen first.

larsrh:
Sorry to see you go!

PR #248, opened 2019-03-02
Change NEScala link from Twitter post to https://nescala.io

bmc:
I think it's better to point to the NEScala web site, now that it's up. The info on the site is more detailed than the info on the Twitter post.


larsrh:
Thanks!

PR #250, opened 2019-04-25
Add cats-mtl; move mainecoon to cats-tagless

rossabaker:


kailuowang:
thanks for updating Mainecoon.

PR #254, opened 2019-04-26
added a donate button

kailuowang:


kailuowang:
sorry @larsrh  I forgot the donate button. added in this PR.

PR #255, opened 2019-04-26
fix a bug in sustainability program announcement

kailuowang:


kailuowang:
merged as this is a straightforward small fix.

PR #257, opened 2019-04-28
Transition to Scala Code of Conduct

rossabaker:
1. Deprecate the Typelevel Code of Conduct. It's left at the original URL, but with an explanation of why we're moving.

2. Add Scala Code of Conduct at `/code_of_conduct.html`.  The only changes are the Domain and the Contacts, in favor of our own to share the load.

3. Replace the Typelevel Code of Conduct with the Scala Code of Conduct in `/about.html`


about.html
```diff
@@ -31,14 +31,14 @@ <h4>Spreadshirt Shop</h4>
       </a>
   </div>
 
-  <h2>Code of Conduct</h2>
-  <p>{% include conduct_header.html %}<br><br>
+  <h2>Scala Code of Conduct</h2>
+  <p>{% include code_of_conduct_header.html %}<br><br>
 
   <a class="btn text center js-expand-coc" href="/conduct.html">Read full Code of Conduct</a></p>
```
larsrh:
```suggestion
  <a class="btn text center js-expand-coc" href="/code_of_conduct.html">Read full Code of Conduct</a></p>
```

rossabaker:
Oops, I overlooked Lars' suggestion, which is correct.  Subsequent PR coming.

SethTisue:
there are some proposed CoC changes here: https://github.com/scala/scala-lang/pull/1079. they are intended to be minor/uncontroversial, but y'all might want to check it regardless

PR #258, opened 2019-04-29
Link to Scala CoC instead of Typelevel CoC from /about.html

rossabaker:
@larsrh caught this on #257 and I overlooked it until after I'd hit merge.


milessabin:
LGTM.

non:
👍

PR #263, opened 2019-05-30
added a banner to promote  fundraising campaign

kailuowang:
Renders like this
Screen Shot 2019-05-29 at 10 05 13 PM


kailuowang:
Thanks @larsrh

PR #269, opened 2019-10-29
Sort list of incubator projects

vlovgr:
- Change to sort list of incubator projects by title.
- Update description for incubator project Ciris.


larsrh:
Thanks!

Issue #272, opened 2019-12-21
Curated projects


Slakah:
This is partly prompted by some thoughts after my [tweet](https://twitter.com/JRCCollier/status/1207808806173990912) in response to the question of how to improve Scala adoption.

> Trusted cohesive library ecosystem. A lot of library usage is built on trust, most folks have been bitten at one points by a lib with some gotchas. This is why I think building trust and branding of libs with shared ethos is very important.

It feels like the [typelevel projects page](https://typelevel.org/projects/) could benefit by curating the functionality provided by the libraries in question. An example of the type of curation I think could be beneficial would be https://github.com/lauris/awesome-scala or https://github.com/rust-unofficial/awesome-rust, in that libraries are grouped based on functionality provided.

An example hierarchy could be:

```yaml
"Web Server":
  - http4s
  - Finch
"HTTP Client":
  - fetch
  - Hammock
  - http4s
"JSON":
  - "argonaut-shapeless"
"Functional Reactive Programming":
  - "cats-effect"
  - "fs2"
  - "monix"
Database:
  - "doobie"
"Front End":
  - "OutWatch"
Configuration:
  - "Ciris"
  - "Extruder"
  - "PureConfig"
Serialization:
  - "scodec"
"Science and Data Analysis":
  - "Algebird"
  - "coulomb"
  - "Frameless"
  - "Libra"
  - "spire"
  - "squants"
"Testing":
  - "cats-scalatest"
  - "ScalaCheck"
  - "scalacheck-shapeless"
  - "Specs2"
  - "discipline"
"Linked Data":
  - "banana-rdf"
"Generic Programming":
  - "shapeless"
"Command Line":
  - "decline"
"Extensions":
  - "cats"
  - "cats-tagless"
  - "eff"
  - "kittens"
  - "monocle"
  - "mouse"
  - "refined"
  - "singleton-ops"
"Misc":
  - "algebra"
  - "paiges"
  - "scala-steward"
  - "scala-exercises"
  - "Scoverage"
"Macros and Compiler Plugins"
  - "export-hook"
  - "imp"
  - "kind-projector"
  - "Machinist"
  - "macro-compat"
  - "simulacrum"
  - "TwoTails"
```

Hopefully this would make it easier to navigate the typelevel ecosystem and find the correct library for the job.

The following projects have been removed, based on lacking recent commits, or being known to be unsupported:

* Alleycats - redirects to cats.
[...]

PR #274, opened 2020-02-29
Curated projects #272

Slakah:
In progress #272 . 

TODO list:

- [ ] Update ruby deps
- [ ] Single yaml typelevel project list
- [ ] Tidy up styling
- [ ] Handle project descriptions which use emojiis (i.e. scala-steward has `:robot:`)
- [ ] Work out what to do with the high profile projects on the homepage
- [ ] What's the best way to keep github stars up to date, period building pipeline?

[In progress preview](https://user-images.githubusercontent.com/16853/75614750-d6988180-5b33-11ea-9ebd-9179bfbc4569.png)


rossabaker:
We'll need to include catbird from #282.

rossabaker:
And case-insensitive from #283

Slakah:
Just to note down an update on where I'm at with this. I mainly lost interest as it was a PITA working with Jekyll, switched focus for the time being, but hope to pick this up again at some point in the not to distant future.

rossabaker:
If you're struggling with the Ruby dependency management, I have had success with the Docker container.  If you're struggling with the Jekyll itself, well, you and me both.  In any event, we appreciate it.

rossabaker:
Further on jekyll: any time I've proposed alternatives to Jekyll on various Typelevel projects, people stand up for how easy it is to deploy on gh-pages.  If we were to do a more radical overhaul, we'd need an answer to that.

Paradox has been my favorite static site generator for the JVM (i.e., no extra install), but I think it's more geared toward docs than blogs.  I use and like Hugo on some other sites, but there would be CI work, and the syntax is less familiar to people.

But that's all probably discussion for another ticket.

PR #276, opened 2020-04-03
Bump versions in Gemfile.lock

valencik:
Trying to build the site today resulted in the following error:

```
Your lockfile is unreadable.
Run `rm Gemfile.lock` and then `bundle install` to generate a new lockfile.
rake aborted!
```

I've followed that advice and these changes are the result of running `bundle install`


larsrh:
Good catch! We should probably test that in CI.

PR #280, opened 2020-04-27
Bump Docker dependencies

rossabaker:
* openjdk-8-jdk isn't available on buster
* sbt-1.0.4 fails to install now, but sbt-1.3.0 works


larsrh:
We've had exactly the same problem with this base image at work: https://www.innoq.com/en/blog/choose-your-docker-base-image-wisely/

rossabaker:
*cough* Nix *cough*

PR #293, opened 2020-10-28
Concurrency in ce3

RaasAhsan:


posts/2020-10-30-concurrency-in-ce3.md
```diff
@@ -0,0 +1,510 @@
...
+contending fibers. Let's take a look at an example.
+
+```scala
+import cats.effect.{IO, IOApp, ExitCode}
+import cats.implicits._
+
+object ExampleFour extends IOApp {
+  override def run(args: List[String]): IO[ExitCode] =
+    for {
+      state <- Ref.of[IO, Int](0)
```
ahjohannessen, line 275: import of `Ref` missing?

SystemFw: change to `IO.ref(0)` actually

djspiewak:
I should have done my comments in-line. Sorry! Copying from Gitter:

- I think we should prefer IOApp.Simple whenever possible in snippets, since it’s more concise and doesn’t involve ExitCode
- I think we should encourage `import cats.syntax.all._` instead of `import cats.implicits._`
- "Throughout its lifetime, a fiber will migrate between these pools to accomplish its various tasks" this is GLORIOUS
- You have exercises?! Damn

RaasAhsan:
* Addressed review comments
* Added some coloring here and there
* Added tut annotations

I think this is otherwise ready, but I'll probably continue making edits in the meantime. Is there anything else I need to get this merged?

arosien:
> I should have done my comments in-line. Sorry! Copying from Gitter:
> 
> * I think we should prefer IOApp.Simple whenever possible in snippets, since it’s more concise and doesn’t involve ExitCode
> * I think we should encourage `import cats.syntax.all._` instead of `import cats.implicits._`
> * "Throughout its lifetime, a fiber will migrate between these pools to accomplish its various tasks" this is GLORIOUS
> * You have exercises?! Damn

I don't understand the point about imports. It's been my understanding that `import cats.implicits._`, along with `import cats._`, is the easiest way to get everything you need. Why the shift?

SystemFw:
> It's been my understanding that import cats.implicits._

`cats.implicits._` imports two sets of things: instances of the various typeclasses, e.g `Monad[Option]`, and extension classes that add syntax, e.g `yourList.foldMap`, or in other words `implicits` is the equivalent of having `instances.all._` and `syntax.all._`. However, recent versions of cats have moved the instances to the companion objects of the typeclasses, so they are always in scope, and all you need is `import cats.syntax.all._` to get the syntax extension

arosien:
> > It's been my understanding that import cats.implicits._
> 
> `cats.implicits._` imports two sets of things: instances of the various typeclasses, e.g `Monad[Option]`, and extension classes that add syntax, e.g `yourList.foldMap`, or in other words `implicits` is the equivalent of having `instances.all._` and `syntax.all._`. However, recent versions of cats have moved the instances to the companion objects of the typeclasses, so they are always in scope, and all you need is `import cats.syntax.all._` to get the syntax extension

Old habits die hard.... :)

Aside: I wonder if `import cats.extensions._` would be better, and hint at Scala 3 terminology. Also the `.all` seems redundant.

djspiewak:
@larsrh can we merge this?

larsrh:
Sure, go ahead.

RaasAhsan:
Pushing a few last minute edits for this, will comment again after it's ready

RaasAhsan:
actually ready now :)

Issue #298, opened 2021-04-09
Should cats-parse be listed in Projects?


RyanHeins:
I just learned about [cats-parse](https://github.com/typelevel/cats-parse) recently, and noticed it's not listed as a [typelevel project](https://typelevel.org/projects/). Should it at least be added to "Incubator Projects", or is there a process to getting added?

larsrh:
Yeah, good idea. Do you want to go ahead with a PR?

rossabaker:
We fell a bit behind on the list while #274 was underway, and that stalled. It would be nice to stay current until that's done.

Log4cats, Vault, Unique (nearly obsolete), Keypool, and as of yesterday fs2-grpc are also missing.

RyanHeins:
I can make a PR to add those. Are they all full-fledged projects, or are any incubating?

rossabaker:
I'd say they're all incubating.  (This is a distinction we'd like to eliminate, but that's most consistent with the status quo.)

Thanks so much for volunteering!

PR #300, opened 2021-05-05
Add incubator projects

RyanHeins:
Also updates readme about adding projects.

This PR addresses issue #298


rossabaker (approved):
Thanks for fixing this!

PR #309, opened 2021-08-24
Add missing double-quote

hmrclt:
Fix for a minor typo


rossabaker (approved):
Thanks!  I'm not sure why CI didn't catch this.

rossabaker:
Oh. I think we don't test it with CI!

PR #313, opened 2021-11-11
renamed joinAndEmbedNever to joinWithNever in 2020-10-30 blog post

josephhajduk:
renamed joinAndEmbedNever to joinWithNever as per https://github.com/typelevel/cats-effect/issues/1500

did not check the correctness of any other examples


rossabaker (approved):
Thanks!  It was correct when it was written, but I don't see any reason not to refresh it for people who are still following it.

Issue #251, opened 2019-04-25
Website build fails even inside Docker container.


tpolecat:
I can't build the website locally or via the docker container. I get `You must use Bundler 2 or greater with this lockfile.`

larsrh:
After poking around I now vaguely remember what the problem may be. Try the following: in the `Gemfile.lock`, remove the last two lines that say essentially "bundled with 2". This should do the trick.

rossabaker:
Docker is gone.  The devshell is a solution for those of us who don't know how to Ruby.

PR #319, opened 2022-01-18
Add a devshell and CI

### LINKED ISSUE #310: Add CI
I broke the jekylls because nothing is testing the site build.  Would be nice to make sure the site builds on PRs.

rossabaker:
Bundler still doesn't work for me, and the Docker container is gone.  This works well for me:

* README.md documents it
* flake.nix drives it
* gemset.nix and flake.lock are generated.  Sorry.

Adds a CI workflow that verifies that the site builds.  Depends on Nix, but I don't care to faff about with Ruby-in-YAML.

Fixes #310.  Mitigates #251.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,23 @@
+name: nix-flake-check
+
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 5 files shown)


rossabaker:
Okay, I don't know why the action won't run.  Maybe something needs to be on the default branch first.

tpolecat (approved):
:+1: this seems good

rossabaker:
I can make it faster if I set up some Cachix keys, but this isn't that slow, so not a priority.

PR #334, opened 2022-04-02
Update 2022-04-01-call-for-steering-committee-members.md

jarrodu:
Fix email link. It did not render correctly.


rossabaker (approved):
Oops.  Thank you!

PR #337, opened 2022-04-18
Update flake

rossabaker:


rossabaker:
I think it's this: https://github.com/NixOS/nixpkgs/issues/158076

There's no urgency to fixing this.  If someone is passionate about C++ and Nix and Darwin, have at it, but I'm 1-for-3.

rossabaker:
Need to solve this to get past the warning:

```
Warning: the running version of Bundler (2.2.24) is older than the version that created the lockfile (2.3.9). We suggest you to upgrade to the version that created the lockfile by running `gem install bundler:2.3.9
```

But #361 fixes it for reboot, so maybe we can let this one go.

rossabaker:
Yeah, I'm over this one.

PR #359, opened 2022-07-22
Fix CWE-190

rossabaker:
The key is to get commonmarker >= 0.23.4.  Fixes https://github.com/typelevel/typelevel.github.com/security/dependabot/2.


rossabaker:
This merge to reboot is going to be a bad time.

Issue #347, opened 2022-07-19
Sorting of projects


rossabaker:
The "above the fold" projects are alphabetical.  Defining "core" projects is tough, but we should do something.

rossabaker:
I just used the existing definition of "core" to select which ones are on the home page.

PR #362, opened 2022-07-22
Restore "core" projects to front page

rossabaker:
We can quibble about which these are, but this matches the status quo.

See #347.


jducoeur:
Yep.  I'll want us to think about what belongs here, but getting back to status quo is the right move IMO.

rossabaker:
I'm merging what I perceive to be non-controversial ones aggressively.  Unless someone says whoa, this one's getting merged in the morning.

Issue #376, opened 2022-07-22
Dynamic cards misbehave on android in Firefox


samspills:
The reboot branch looks great in desktop Firefox, or desktop view in Firefox on android, and content loads as I scroll down the site.

However mobile view in Firefox doesn't display the dynamic content as I scroll down. When I hit the bottom of the page and start scrolling back up I see only 1 blog card, but all the project and goal cards appear. Once I hit the top and start scrolling back down everything is gone again.

Here's a screen recording of the current `reboot` behaviour:

https://user-images.githubusercontent.com/4977741/180506132-827694d4-8778-43d1-aecf-ef866ce6d17f.mp4

rossabaker:
Eek.  I don't see this in Firefox on the desktop, but can reproduce it using the Responsive Design Mode in its Web Developer Tools.  Fixing it is probably going to require someone frontendier than me.

samspills:
Cc @raulraja and @israelperezglez as Raul requested in the discord :)

israelperezglez:
Thank for reporting it. I will fix it ;)

PR #378, opened 2022-07-24
Fix README typo and add 2022-07 new members post

JasnaMRB:
Slightly changed closing paragraph and and opening paragraph from steering committee's shared draft -- added a reference to the previous post in the opening paragraph.

I'll be out of town this upcoming week, so if any changes need to be pushed, @rossabaker or others are welcome to do so! I should still have access to comment on this PR and press buttons via my phone, though. I set the date to this coming Monday for now -- open to that change as well.

FYI: I can't get the site to build via the first recommended method in the README, so I haven't been able to preview these changes other than via my Markdown viewer. Unfortunately I don't have time to look too deep into it yet...

gist in case it may be worth a separate bug report, or otherwise I may have missed a build step: https://gist.github.com/JasnaMRB/8600a578865596cdb56ed25bf2eb4d8b

Issue #377, opened 2022-07-23
Right side of site too far right on Android chrome


JasnaMRB:
The main content of the homepage is appropriately appearing scrolling up and down, but the hamburger menu on the right isn't immediately visible unless i scroll a few strokes to the right since the computer + person splash image takes up space. See screenshots below, sorry i can't figure out how to take video with me scrolling at the same time!

cc @raulraja @israelperezglez

Zoomed out view
![SmartSelect_20220722-213133_Chrome.jpg](https://user-images.githubusercontent.com/568975/180585340-eb6e9fd2-734e-40c2-9b4c-3ed0e1f2cfe2.jpg)

Hamburger menu after scrolling all the way right
![Screenshot_20220722-213124_Chrome.jpg](https://user-images.githubusercontent.com/568975/180585343-55796198-d689-4da1-a424-6f08eb4592b0.jpg)

Initial view when loading the screen on Android chrome
![Screenshot_20220722-213111_Chrome.jpg](https://user-images.githubusercontent.com/568975/180585353-e041a9cd-7b84-4046-b73c-6d6b10e46658.jpg)

rossabaker:
Good find.  This one reproduces in developer tools, as well.

israelperezglez:
Thank for reporting it. I will fix it ;)

PR #387, opened 2022-08-16
Revision responsive mobile Typelevel site

israelperezglez:
- ✅ We have created a placeholder in the post that is shown when there is no image defined in the image variable of the post.
 - ✅ Fixed the display of the tag in the list of libraries when a library does not have one.
 - ✅ Reviewed the responsive of the home page. We have a css clip-path property that in mobile we had to edit and change the way to use it.
 - ✅ Reviewed the responsive of the blog filter
 - ✅ Reviewed the general responsive of the css grid we use as wrapper.
 - ✅ Reviewed the responsive of the h1 and the path we use in the pseudo selectors ::after.
 - ✅ Reviewed the responsive of the about section and corrected the margins of the .about-item class.
 - ✅ We have decreased the loading time of the AOS library that we use in the different elements when we scroll.


rossabaker (commented):
Thanks!  Confirmed locally this fixes #376 and #377.

Both this one and #378 change the placeholder image.  We'll need to decide which one we want.

rossabaker:
I don't have rights to push to this branch, but I resolved the merge conflicts on issue-387 branch.  There will be more after #388.

rossabaker:
Actually, it might be easier to just merge this into @JasnaMRB's PR... and then everything will launch together.  I'll post screenshots of the placeholders over there so we can decide.

rossabaker:
This was folded into #388.  Many thanks!

PR #389, opened 2022-08-18
remove Typelevel Scala from projects list

SethTisue:
it's dead/archived

### DIFF
--- a/collections/_projects/scala.md
+++ b/collections/_projects/scala.md
@@ -1,9 +1,0 @@
----
-layout: post
-title: "Scala"
-category: "Our fork of the Scala compiler"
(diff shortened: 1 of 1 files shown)


jducoeur (approved):

johnynek (approved):

rossabaker:
There are actually a lot of archived projects.  I marked them all in the front matter with `archive`, but we're not doing anything special to render them.  I wasn't sure if we wanted to keep them for historical interest or just banish them.

SethTisue:
> I wasn't sure if we wanted to keep them for historical interest

it's not worth making life more difficult for newcomers and casual users, IMO

rossabaker:
In that case, there are [several more](https://github.com/typelevel/typelevel.github.com/search?q=archived).

SethTisue:
> In that case, there are several more

#390

PR #398, opened 2022-09-23
Fix #392: Add blog/feed.rss

### LINKED ISSUE #392: RSS feed for blog
Calling this a "bug" is a bit borderline, but it's a regression from the previous version so I think it's appropriate.  The blog should have an RSS feed, preferably at the same URL (https://typelevel.org/blog/feed.rss) as the old one.

It isn't clear offhand where the old feed came from.  Was it generated in post-processing?  Did it somehow come out of the server?

alexandru:
I tested locally, discovery and loading the RSS file in a client works.

File copy/pasted from the old website design. My website's own RSS file does more work and sanitization than this, but this is fine for now.

### DIFF
--- a/_includes/_head.html
+++ b/_includes/_head.html
@@ -14,5 +14,6 @@
   <link href="https://unpkg.com/aos@2.3.1/dist/aos.css" rel="stylesheet">
   <link rel="stylesheet" href="{{ '/css/main.css' | relative_url }}">
 
+  <link rel="alternate" type="application/rss+xml" title="Blog articles" href="{% link blog/feed.rss %}" />
   
 </head>
--- a/blog/feed.rss
+++ b/blog/feed.rss
@@ -1,0 +1,26 @@
+---
+layout: empty
+---
+<?xml version="1.0" encoding="UTF-8"?>
+<rss version="2.0" xmlns:atom="http://www.w3.org/2005/Atom" xmlns:dc="http://purl.org/dc/elements/1.1/">
+	<channel>
+		<title>{{ site.name | xml_escape }}</title>
+		<description>{% if site.description %}{{ site.description | xml_escape }}{% endif %}</description>
+		<link>{{ site.baseurl }}</link>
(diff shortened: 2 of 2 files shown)


jducoeur (approved):

rossabaker:
Thanks for bringing this back.  I'm trying to get back into RSS myself...

alexandru:
Thanks for merging 🙂 I noticed it got merged when my feed reader started seeing new posts again.

Screenshot 2022-09-24 14 34 21

---

@rossabaker if you want to get back to RSS, here are some recommendations:

* If you're looking for a managed service, [Newsblur](https://newsblur.com/) is the best I've seen;
* If you're into self-hosting (which is what I do), [FreshRSS](https://freshrss.org/) is pretty painless via Docker;
* If on macOS or iOS, [Reeder](https://reederapp.com/) is a proprietary client that's been well maintained (can connect to FreshRSS or Newsblur, too);
* Android doesn't have good options for 3rd party clients, but Newsblur's Android client is decent, and FreshRSS is usable from the browser;

You can also follow Twitter feeds:
- Newsblur can do it out of the box;
- I self-host, so I use [RSS-Bridge](https://github.com/RSS-Bridge/rss-bridge);
- I also have a "Subscribe to RSS" and "Subscribe to Twitter/RSS" bookmarklets that I wrote myself (nothing much, it's just a piece of JS that opens a popup with the right URLs and form values), since RSS feed discovery can be challenging, and available browser extensions suck;

When going through your RSS feed, it helps if you also have a read-it-later service or system, since you notice that some articles are good, but don't have time for them. I use [Pocket](https://getpocket.com/), but you can use some sort of TO-DO app as well (e.g., CalDAV / org-mode).

jducoeur:
I know that this only affects a small number of people, but for the Dreamwidth users out there, the blog is available as [typelevelblog-feed](https://typelevelblog-feed.dreamwidth.org/).  (I probably set that up, years ago.)

PR #400, opened 2022-09-30
Update site baseUrl to fix RSS feed permalinks

keynmol:
This is what the links on the RSS feed look like now:

```
		<link>/typelevel.github.com</link>
			<link>/typelevel.github.com/blog/2022/09/19/typelevel-native.html</link>
			<link>/typelevel.github.com/blog/2022/09/12/tuple-announcement.html</link>
			<link>/typelevel.github.com/blog/2022/09/06/new-website-layout.html</link>
			<link>/typelevel.github.com/blog/2022/07/25/welcoming-new-steering-committee-members.html</link>
			<link>/typelevel.github.com/blog/2022/04/01/call-for-steering-committee-members.html</link>
			<link>/typelevel.github.com/blog/2022/01/19/governing-documents.html</link>
			<link>/typelevel.github.com/blog/2021/11/15/on-recent-events.html</link>
			<link>/typelevel.github.com/blog/2021/05/05/discord-migration.html</link>
			<link>/typelevel.github.com/blog/2021/04/27/community-safety.html</link>

```

Which doesn't play well with RSS viewers that load the page in an embedded browser.

You can verify this by `curl https://typelevel.org/blog/feed.rss | grep '<link>' | head`

The link is used here: https://github.com/typelevel/typelevel.github.com/blob/development/blog/feed.rss#L21

Note that I haven't built the site so I'm not 100% sure this is a safe change..

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -16,7 +16,7 @@ permalink: pretty
 
 # Serving, url
 # ---------------------------------------------------------------
-baseurl: "/typelevel.github.com"
+baseurl: "https://typelevel.org"
 url: ""
 
 # Conversion


rossabaker (approved):
Seems right.  I'll test after publish.

PR #401, opened 2022-10-01
Revert "Update site baseUrl to fix RSS feed permalinks"

djspiewak:
Reverts typelevel/typelevel.github.com#400

Breaks the site. Let's roll this back while we fix.


samspills (approved):

rossabaker:
Sorry.  It didn't deploy until long after the action completed, so I went to bed.

PR #404, opened 2022-11-01
Add a social pointer to Mastodon

### LINKED ISSUE #403: Link to Mastodon with rel=me
This will allow verification of the profile.  More detailed instructions are on the profile page.  Do this before setting the website there.

jducoeur:
Specifically, our new official home at Fosstodon. This also serves as a verification link for Fosstodon, and the `rel="me"` is required, which is why it's being handled out-of-band from the other links; if we're comfortable always using `rel="me"` for social links, we could unify this.  (Opinions welcome.)

Fixes #403.

### DIFF
--- a/_data/nav-social.yml
+++ b/_data/nav-social.yml
@@ -6,6 +6,10 @@
   url: https://twitter.com/typelevel
   icon: "fab fa-twitter"
 
+- title: Mastodon
+  url: https://fosstodon.org/@typelevel
+  icon: "fab fa-mastodon"
+
 - title: Discord
   url: https://discord.gg/XF3CXcMzqD
   icon: "fab fa-discord"
--- a/_data/nav.yml
+++ b/_data/nav.yml
@@ -23,6 +23,10 @@ navSocial:
     url: https://twitter.com/typelevel
     icon: "fab fa-twitter"
 
+  - title: Mastodon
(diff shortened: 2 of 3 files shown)


rossabaker (approved):
[`rel=me`](https://microformats.org/wiki/rel-me) is useful when it's bidirectional, and harmless when it's unidirectional.  It would be appropriate to put on all the links, and then any services that link back would establish that verified relationship.

Concrete example: On https://indieweb.social/web/@rossabaker, Mastodon added `rel=me` to both my personal site and my Matrix handle.  rossabaker.com links back to Mastodon with a `rel=me`, so it's green.  matrix.to doesn't, and I can't make it do so, so it's not green.  Now you know that Mastodon is the same person as rossabaker.com, and if you trust one, you can trust both.   But we could both be pointing to someone else's Matrix, because Matrix doesn't link back.

samspills (approved):
Thank you 🎉

JasnaMRB (approved):

jducoeur:
Okay -- I think Ross and I are thinking along similar lines, so I'm going to rewrite this to do `rel="me"` everywhere.

jducoeur:
Updated.  Folks should give it a glance, just to make sure we're on the same page.

rossabaker (approved):

samspills (approved):

PR #412, opened 2023-01-04
Add filter tabs to Projects page

valencik:
Add `/projects/organization` and `/projects/affiliate` pages along with tab selectors.

All Projects:
allProjects

Organization Projects:
organizationProjects

Afilliate Projects:
affiliateProjects

### DIFF
--- a/_data/filter-projects.yml
+++ b/_data/filter-projects.yml
@@ -1,0 +1,9 @@
+- title: All Projects
+  url: /projects/
+  category: allProjects
+- title: Member
+  url: /projects/member/
+  category: member
+- title: Affiliate
+  url: /projects/affiliate/
+  category: affiliate
--- a/_data/filter-projects.yml
+++ b/_data/filter-projects.yml
@@ -1,9 +1,9 @@
 - title: All Projects
   url: /projects/
   category: allProjects
-- title: Member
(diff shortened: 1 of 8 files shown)


_data/filter-projects.yml
```diff
@@ -0,0 +1,9 @@
+- title: All Projects
+  url: /projects/
+  category: allProjects
+- title: Member
+  url: /projects/member/
```
rossabaker, line 5: We've called these Organization projects in the charter and in the proposed badges.

rossabaker (commented):
I wish the templates weren't so redundant, but that seems to be the way of things in this repository, and I don't have the Jekyll chops to do better.  I like the feature and think we should move forward.

_data/filter-projects.yml
valencik: oh oops, good catch

valencik:
The shields proposed in https://github.com/typelevel/governance/issues/72 do include the word `project`, should we stick with exactly that? Probably...

rossabaker (approved):
I think "project" on the badges is good because "Typelevel | Organization" looks weird.

zmccoy (approved):

samspills (approved):

PR #418, opened 2023-01-25
Consolidate project data to one yaml file

valencik:
This PR consolidates all the project data to one yaml files as done in https://github.com/typelevel/typelevel.github.com/pull/274 but without the categorization, which could perhaps come later in a separate PR.

It also makes sure cats comes first in the list of `cats-...` named projects.

Screen Shot 2023-01-24 at 9 44 19 PM

I've kept changes to a minimum here, but I feel like this would make it easier to maintain reasonable `category` values across all the projects. Or ensure that all projects have particular metadata.


valencik:
Just a quick note to self, if we go with this approach we'll want to update the migration template: https://github.com/typelevel/governance/blame/main/.github/ISSUE_TEMPLATE/2-project-migration.md#L22

zmccoy:
I think the single file is fine just ease of ordering etc.  We can just keep an eye on if this becomes a pain.

zmccoy (approved):

samspills (approved):
Also a fan of the single file, and yah not impossible to chunk back up into individual files if we want that in the future

PR #430, opened 2023-03-06
Setup dependabot

armanbilge:
This will keep the actions up-to-date, which would have prevented the current CI failures.


rossabaker (approved):
In my experience, this runs shortly after merge, so merging it now should get us the PRs that will get it to green.

PR #443, opened 2023-04-22
Add Natchez, Skunk, fs2-compress projects

valencik:
Adds Natchez and Skunk as per https://github.com/typelevel/governance/issues/94

Adds fs2-compress as per https://github.com/typelevel/governance/issues/90


_data/projects.yml
```diff
@@ -240,6 +248,9 @@
   description: "Operations for primitive and String singleton types"
   github: "https://github.com/fthomas/singleton-ops"
   affiliate: true
+- title: "Skunk"
```
samspills, line 251: Should we add the same category tag as doobie? "Principled database access"?

samspills (approved):

_data/projects.yml
valencik: I was thinking about that as well, but then thought the whole list of categories could use a review and didn't.
I might even steal the category label and use it for denoting platform support. I don't think the category label is carrying its weight.

jducoeur: Probably worth a conversation about what our taxonomy ought to be, yes.

PR #457, opened 2023-06-29
Use github-pages gem

kiendang:
Sometimes the deployed site looks different compared to one built locally. This is due to Github Pages using a different version of `jekyll` than the one currently defined in Gemfile. See https://pages.github.com/versions/. Can see `Warning:  github-pages can't satisfy your Gemfile's dependencies.` in the [deploy action](https://github.com/typelevel/typelevel.github.com/actions/runs/5414574615/jobs/9841829888).

This uses the [`github-pages`](https://github.com/github/pages-gem) gem instead of fixing a `jekyll` version to keep the locally built site in sync with the deployed site.


rossabaker:
I think this build might fail on the Nix derivation, but we can clean that up.

I like the approach.  If this doesn't work, we could stop letting GitHub run Jekyll and run our own Jekyll and just push the rendered site.  Then it's impossible to get out of sync.

kiendang:
> If this doesn't work, we could stop letting GitHub run Jekyll and run our own Jekyll and just push the rendered site. Then it's impossible to get out of sync.

I do like this, having more control over what gems and versions to use. Looks like the `github-pages` gem pulls in a bunch of unnecessary dependencies.

Also I'm not sure how github-pages works. Like does it respect `Gemfile.lock` at all or always use the latest version of the `github-pages` gem? Coz if it's the later then it would be hard to keep local dev env in sync.

rossabaker:
That's a good question.  Unless it specifically reads github-pages out of our Gemfile.lock, I don't think it would respect our build any more than it currently is.  It would just give us the means to get a build closer to theirs... if we could keep it synchronized with their releases.

An alternative here is to run our own pinned Jekyll in a publish action and push the rendered site to a branch with a `.nojekyll` file.  I think this is how sbt-microsites works.  But I'm willing to try this approach and see how hard that synchronization is in practice.

kiendang:
Now I'm pretty sure Github Pages will just use the latest version of the github-pages gem. One way to keep the dev env in sync is to set up dependabot to automatically bump the github-pages version in Gemfile.lock to the latest version. However if I understand correctly gemset.nix needs to be regenerated whenever Gemfile.lock changes? Would need to either manually regen gemset.nix whenever there's a new dependabot PR or set up another action for that.

PR #468, opened 2023-10-02
Fix NE Scala URL

jducoeur:
The content is currently pointing to an illegal version of the URL.  Maybe the publication problem is that the Markdown generator is silently checking URLs, and this is getting a 404?  Let's try changing to a known-good that doesn't involve any redirection, and see if that fixes it.


samspills (approved):
worth a shot!

PR #476, opened 2023-11-27
Add cats-stm to projects

valencik:
Adding cats-stm as an affiliate project as per https://github.com/typelevel/governance/issues/119

cc. @TimWSpence


rossabaker (approved):

TimWSpence:
Awesome, thanks @valencik! :)

Issue #420, opened 2023-02-06
Projects with no `affiliate` label show as organization projects


valencik:
Consider cats-time

```yaml
- title: "Cats-Time"
  description: "Instances for Cats Typeclasses for Java 8 Time"
  github: "https://github.com/ChristopherDavenport/cats-time"
```
<img width="373" alt="Screen Shot 2023-02-06 at 4 11 38 PM" src="https://user-images.githubusercontent.com/5440389/217089025-d8ccabd8-29a4-443a-adf1-01a198904c4a.png">

armanbilge:
Or maybe it's just smarter thank you think 😜 

https://github.com/typelevel/cats-time/

valencik:
It remains the case that project membership is defined with a non required `affiliate` boolean.

```
<div class="project-membership">
  {% if project.affiliate %}
  <img src="https://img.shields.io/badge/typelevel-affiliate%20project-FFB4B5.svg" alt="Typelevel Affiliate Project" />
  {% else %}
  <img src="https://img.shields.io/badge/typelevel-organization%20project-FF6169.svg" alt="Typelevel Organization Project" />
  {% endif %}
</div>
```

I wonder if we should instead have an explicit `membership: affiliate` or `membership: organization`?

PR #480, opened 2023-12-27
Remove `affiliate` from a few org projects

valencik:
Some of this was originally done in https://github.com/typelevel/typelevel.github.com/pull/424
But I think I messed things up in the big one file squish.
Oh well, I've gone through all the affiliate projects now and I think this is it.

### DIFF
--- a/_data/projects.yml
+++ b/_data/projects.yml
@@ -315,7 +315,6 @@
 - title: "simulacrum"
   description: "First-class syntax for type classes"
   github: "https://github.com/typelevel/simulacrum"
-  affiliate: true
   platforms: [js, jvm, native]
 - title: "Simulacrum Scalafix"
   description: "Simulacrum as Scalafix rules"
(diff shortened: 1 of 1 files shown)


valencik:
I'm reminded of this annoyance: https://github.com/typelevel/typelevel.github.com/issues/420#issuecomment-1419770016

rossabaker:
Some were flagged as such because they were orginally Incubator projects.  When that went away, nobody had confirmed with the maintainers that they accepted the new terms of the charter, so all Typelevel Organization projects are in the Typelevel GitHub org, but not vice versa.  It would be great if this is reconciled, but that, I think, is why.

valencik:
I was wondering how to reconcile this, and came across the following in the Charter:

> 8.3 Initial projects.** Projects already hosted under the Typelevel GitHub organization when this charter is ratified are considered Organization projects unless they are later removed.

https://github.com/typelevel/governance/blame/227296116f582064bcc33ff63972e06221061647/CHARTER.md#L70

Because we have the option of removing projects in the future, I think this is safe to merge.
Should one of the original maintainers disagree with the Charter rules, the project could be removed.

Pinging the affected teams just as a FYI
@typelevel/kittens @typelevel/simulacrum @typelevel/squants 


I think we should approve this PR and merge to remove any ambiguity in the status of projects.

jducoeur (approved):

rossabaker (approved):

PR #486, opened 2024-01-25
Add WKD with the Typelevel release PGP key

rossabaker:
1. Sets up a [Web Key Directory](https://wiki.gnupg.org/WKD), so the public release key (14CA 4AC4 FDB0 0D5F 827A  BC0C 5986 48A6 5DF6 4506) can be found via `--locate-keys`.
2. Contains a [DNS identity claim](https://docs.keyoxide.org/service-providers/dns/) to mutually link it with typelevel.org.  The `TXT` record already exists.


rossabaker:
The alternative here is to set up an openpgp subdomain and do the "advanced" implementation, whose directory structure looks like the first commit.  That "site" could conceivably be built from public key(s) as the source.

samspills (approved):

rossabaker:
Advanced is a little easier to automate in an action because:
- gpg-wks-client generates that directory structure.
- It's its own thing and doesn't need to be integrated into Jekyll.
- It's basically `(DomainName, Set[PublicKey]) => IO[Victory])`.  [^1]

Direct could be automated by removing an intermediate directory (the oops I fixed in eda22b3a328bf9ed4cdb542166ca6cf53ac0ffed) and by not screwing up the rest of Jekyll.

Clients are advised to consult Advanced first, and fall back to Direct.  

[^1]: Why, yes, I do already have a generic Nix derivation that does this, I'm glad you asked.

rossabaker:
I'm going to merge this to get something working, and we can debate Advanced and automation when we have to update it before it expires again in July.

rossabaker:
Follow-up: this passes all four [direct WKD tests](https://metacode.biz/openpgp/web-key-directory) and [successfully verifies the domain with Ariadne](https://keyoxide.org/noreply%40typelevel.org).

GitHub verification would be neat, but requires a gist, which orgs can't do.  Mastodon and OpenCollective are possible, but really nave nothing to do with this key.

PR #536, opened 2025-01-02
Feral blog post formatting: replace markdown table

samspills:
The markdown table isn't rendering. I'm not sure why and I don't really have the time to dedicate to debug properly, so I'm proposing that section be rewritten as a list


samspills:
Arman figured it out actually, we just need a newline

PR #544, opened 2025-02-19
Create security.txt

rossabaker:
Follows https://www.rfc-editor.org/rfc/rfc9116

The key is signed by the same PGP key that signs our artifacts and is published to our site, but the referenced encryption keys are mine and @armanbilge's, as they appear on the Typelevel security policy.  This distinction makes sense in my mind, because Arman and I don't typically sign artifacts for Typelevel, and the Typelevel bot shouldn't receive encrypted messages.

This is something that should be renewed every year.

/cc @typelevel/security

### DIFF
--- a/.well-known/security.txt
+++ b/.well-known/security.txt
@@ -1,0 +1,17 @@
+-----BEGIN PGP SIGNED MESSAGE-----
+Hash: SHA512
+
+Contact: mailto:security@typelevel.org
+Expires: 2026-02-19T02:39Z
+Encryption: openpgp4fpr:904c153733dbb0106915c0bd975be5bc29d92ca5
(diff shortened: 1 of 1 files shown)


jducoeur (approved):
Makes sense, and appears to match the RFC.

Does raise the question of how we track and remember to renew things like this (when there isn't a company dunning us with email reminders), but that's not a blocker.

rossabaker:
If we remember to renew it when we renew the GPG key that signs it, they're on approximately the same cycle.

We could also establish a shared key for the Security Team, which would simplify some things and complicate others, but that's something that can be done any time after this.  I'd just like to have something reasonable in place as a starting point.

PR #551, opened 2025-03-03
Remove doobie project idea

armanbilge:
Unfortunate oversight, it seems like this work was already completed but the issue was not closed 😕 

- https://github.com/typelevel/doobie/issues/2147#issuecomment-2693444127


jatcwang:
There's still quite a lot of tests with unsafeRunSync so the ticket is not done yet! :)

armanbilge:
Thanks for clarifying!

PR #562, opened 2025-08-17
Dashed off a quick post celebrating the new submarine error handling

djspiewak:
This should be held until we do the next Cats MTL release (today?) and the post should be updated to reflect the relevant version number. Also could use a quick glance to make sure I didn't completely botch the syntax and semantics. *poke* @lenguyenthanh


collections/_posts/2025-08-17-custom-error-types.md
```diff
@@ -0,0 +1,150 @@
...
+category: technical
+
+meta:
+  nav: blog
+  author: djspiewak
+---
+
+One of the most famous and longstanding limitations of the Cats Effect `IO` type (and the Cats generic typeclasses) is the fact that the only available error channel is `Throwable`. This stands in contrast to bifunctor or polyfunctor techniques, which add a typed error channel within the monad itself. You can see this easily in type signatures: `IO[String]` indicates an `IO` which returns a `String` or may produce a `Throwable` error. Something like `BIO[ParseError, String]` would represent a `BIO` that produces a `String` *or* raises a `ParseError`. The latter type signature is more general than `Throwable`, since it allows for user-specified error types, and it's somewhat more explicit about where errors can and cannot occur.
+
+In a meaningful sense, type type of bifunctor error encoding is analogous to *checked* exceptions in Java, whereas monofunctor error encoding (like Cats Effect's `IO`) is analogous to *unchecked* exceptions. Both are valid design decisions for an effect type, but they come with different benefits and tradeoffs.
```
SethTisue, line 13:
```suggestion
In a meaningful sense, this type of bifunctor error encoding is analogous to *checked* exceptions in Java, whereas monofunctor error encoding (like Cats Effect's `IO`) is analogous to *unchecked* exceptions. Both are valid design decisions for an effect type, but they come with different benefits and tradeoffs.
```

collections/_posts/2025-08-17-custom-error-types.md
```diff
@@ -0,0 +1,150 @@
...
+
+However, the basic idea of MTL itself, divorced from the *datatypes* (like `EitherT`), is actually a very good one. At its core, MTL is just about expressing capabilities available within a given scope using implicit evidence. Capabilities can be things like parallelism, resource safety, error handling, dependency injection, sequential composition, or similar. When done correctly, this can be a very powerful and lightweight way of expressing compositional effects with a high degree of granularity and type safety. It's not a coincidence that this is exactly the route being explored by many of the researchers working on Scala academically!
```
SethTisue, line 35: maybe add a link/reference here to at least one of the routes you're referring to...?

lenguyenthanh:
@djspiewak anything I can help to finish this up?

djspiewak:
@lenguyenthanh I seem to be making a habit of leaving things half-finished! I won't have any time to do wordsmithing on this at least until early next week, but if you want to suggest some edits, I'm happy to incorporate them and that might be enough to unstick things! I'm very aligned with the comments you (and others) have left, so there's nothing controversial, just a matter of getting it revised.

djspiewak:
Updated the post! i think this should be ready to go, or at least close to it.

armanbilge (commented):
Some thoughts, great post, let's publish it! 😁

collections/_posts/2025-08-17-custom-error-types.md
```diff
@@ -0,0 +1,152 @@
...
+
+```scala
+def parse(input: String): IO[Either[Failure, Result]] = ???
+```
+
+And then of course, everything you do with that result must be explicitly `flatMap`ped into the `Either`, and higher-order control flow libraries like Fs2 will often need some extra coaxing in order to make everything work the way you want it to. This gets old in a hurry, which often results in reaching for alternatives like `EitherT`. That way lies madness.
+
+## Capabilities
+
+The good news is that we now have a better answer here, and one which composes perfectly with the existing (and future) ecosystem, maintains all relevant concurrency properties, and which type-infers extremely well, particularly in Scala 3. The answer has been to double down on the relatively little-used implicit capabilities library for Cats, known under the very misleading name of Cats MTL.
```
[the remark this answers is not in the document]
djspiewak: lol okay okay. I'll soften it.

I do think there are a couple edges, but they tend to show up in cases where having multiple channels would have been inherently dicy.

collections/_posts/2025-08-17-custom-error-types.md
```diff
@@ -0,0 +1,152 @@
...
+
+Rather than trying to impose a particular multi-channel composition semantic on your code, we simply stick with a single error channel with known and well-understood supremacy semantics, and everything else follows from there.
+
+## Conclusion
+
+Hopefully you find this technique helpful! This has been in the works for a *surprisingly* long time (I think it was first suggested in the Typelevel Discord about two or three years ago), and it was Thanh Le ([@lenguyenthanh](https://github.com/lenguyenthanh)) who ultimately pushed it over the line. Huge shoutout! He has already begun leveraging this functionality in Lichess, one of the larger production Scala projections: [lichess-org/lila#17944](https://github.com/lichess-org/lila/pull/17944)
```
armanbilge, line 148:
```suggestion
Hopefully you find this technique helpful! This has been in the works for a *surprisingly* long time (I think it was first suggested in the Typelevel Discord about two or three years ago), and it was Thanh Le ([@lenguyenthanh](https://github.com/lenguyenthanh)) who ultimately pushed it over the line. Huge shoutout! He has already begun leveraging this functionality in Lichess, one of the larger production Scala projects: [lichess-org/lila#17944](https://github.com/lichess-org/lila/pull/17944)
```

benhutchison:
Thanks for your work on this @djspiewak 🙏 

Nicely written post. Beyond advertising a great new feature, to me the post also:
- Better explains and contrasts the Typelevel design approach compared to ZIO's typed errors.
- Highlights the strong and mature support for Capability traits in the Typelevel approach. I believe that many developers building "tagless-final style" apps on the Typelevel stack routinely enjoy the benefits of capability-traits, to the point we take it for granted. 

I do think we sorely need better syntax for type holes in Scala 3 so we can use context-bound syntax, eg:
```
def parse[F[_]: Raise[*, ParseError] as r: Monad[F]](input: String): F[Result] 
```

benhutchison:
Actually, I'm worried that the example code in the blog post won't actually compile as written 😟 

The use of `raise` postfix as below requires syntax to be imported? 

In my testing it seems I need add `import cats.mtl.syntax.raise.*` at minimum..  

```
// use that error type in some function
def parse[F[_]](input: String)(using Raise[F, ParseError], Monad[F]): F[Result] =
  // do some hardcore parsing
  if missingBracket then
    UnclosedBracket.raise[F, Result]
  else if missingSemicolon then
    MissingSemicolon.raise // we can rely on type inference and omit extra typings
  else
    result.pure[F]
```

PR #579, opened 2026-02-18
mention Scala on landing page and colophon page

SethTisue:
on discord, @jducoeur indicated that some form of the landing page change would be welcome

it's of course bikesheddable exactly where to insert "Scala" and whether it should be a link. here's one possibility

### DIFF
--- a/src/colophon.md
+++ b/src/colophon.md
@@ -1,6 +1,6 @@
 # Colophon
 
-We build this website with our own tooling! It is statically generated with [Laika], a Typelevel project for transforming Markdown into HTML sites and e-books. The search bar is powered by [Protosearch], an in-memory search library with advanced querying features. Both projects use [Cats] and other Typelevel libraries.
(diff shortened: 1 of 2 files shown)


jducoeur (approved):
Makes sense to me, although I'll defer to Arman and Andrew, who have been driving the vision so far.

SethTisue:
Andrew already altered the colophon page to mention Scala further down, but I (mildly) think my change (to also mention it right up front) is desirable. the first mention will serve to emphasize that the whole thing is Scala-based; the later mention then adds that y'all are using Scala for your scripting, too

reardonj:
> My (possibly unfounded) fear is that by mentioning Scala too upfront, potential users may move along because they are not using the Scala programming language.
> 
> Instead, my vision for the website is to entice curious developers with functional programming and compelling examples of how to use our stack, and that they will learn about Scala along the way :)

I could see this working if there was a next page showing the user more specifically what the benefits of using Cats/CE/fs2/etc. Presumably this would include some explanation of why Scala is a specifically good choice for all this. The current site though is pretty cryptic and users really only have the project page to figure out what is going on right now, which has cryptic tags like `js`, `jvm`, `native`.

IMO, it might be helpful for the time being to mention Scala so uninformed visitors (and search engines) have some notion of what is going on.

Issue #149, opened 2017-03-05
Render MathJax in an RSS-reader-friendly way


S11001001:
As noted in #148, readers don't handle MathJax; can we preprocess MathJax blocks while creating the RSS output into something more likely to be useful?

![screenshot_20170305-175848](https://cloud.githubusercontent.com/assets/219186/23589627/71a81322-019e-11e7-85dd-b6292876f2fd.png)

PR #584, opened 2026-02-19
Render math in RSS with MathML

armanbilge:
Closes https://github.com/typelevel/typelevel.github.com/issues/149.

My RSS reader appears to support MathML. To make this work, needed a couple tweaks:
1. When rendering to RSS, disable KaTeX's HTML output in favor of MathML.
2. Because the RSS renderer secretly delegates to the HTML renderer, we need a bit of trickery to get the correct content to render.

### DIFF
--- a/build.scala
+++ b/build.scala
@@ -452,6 +452,7 @@ object KaTeX {
       output: String = "htmlAndMathml"
   ): String =
     synchronized {
+      // https://katex.org/docs/options
       val options = Map(
         "throwOnError" -> true,
         "strict" -> true,


build.scala
```diff
@@ -283,9 +283,9 @@ object LaikaCustomizations {
       SpanDirectives.create("math") {
         import SpanDirectives.dsl.*
         rawBody.map { body =>
-          RawContent(
-            NonEmptySet.of("html", "rss"),
-            KaTeX(body, false)
+          SpanSequence(
+            RawContent(NonEmptySet.of("html"), KaTeX(body, false)),
+            RawContent(NonEmptySet.of("rss"), KaTeX(body, false, "mathml"))
```
valencik: So the idea is roughly:

We register a `@math` directive, usable as a span or block.
It is formatted as a span with to pieces of content ([SpanSequence](https://javadoc.io/doc/org.typelevel/laika-docs_2.12/latest/laika/ast/SpanSequence.html)).
This first [RawContent](https://javadoc.io/doc/org.typelevel/laika-docs_2.12/latest/laika/ast/RawContent.html) only renders if we are targeting the html format, and in that case we use the default of KaTex output which is `htmlAndMathml`.
The second RawContent only renders if we are targeting our custom rss format in which case we use the `mathml` output.

PR #591, opened 2026-02-20
Add a couple of Doodle projects

noelwelsh:
This PR adds a couple of projects from Creative Scala to the GSoC list.

### DIFF
--- a/src/gsoc/directory.conf
+++ b/src/gsoc/directory.conf
@@ -170,7 +170,7 @@ ideas: [
     difficulty: Medium
     length: ${medium-length}
     mentors: [noelwelsh]
-    categories: [graphics]
+    categories: [media, programming languages]
     repolinks: [
       {
         name: Doodle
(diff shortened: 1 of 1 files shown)


src/gsoc/directory.conf
```diff
@@ -161,5 +161,37 @@ ideas: [
...
+  },
+
+  {
+    title: Doodle Immediate Mode Algebra
+    description: "Design and implement an API for Doodle that allows low-level bitmap based operations."
+    prereqs: "Proficient with Scala and some understanding of computer graphics."
+    difficulty: Medium
+    length: ${medium-length}
+    mentors: [noelwelsh]
+    categories: [graphics]
```
armanbilge: The closest GSoC categories are "media" (which includes graphics) and "programming languages" (which includes libraries). Also, I realized I forgot to render the categories in the page, so I need to fix that 😅 

```suggestion
    categories: [media, programming languages]
```

armanbilge (approved):
thank you!

PR #635, opened 2026-04-10
Add WebJar class for accessing WebJar files

valencik:
Closes https://github.com/typelevel/typelevel.github.com/issues/633

There's basically two features of this work:
- We only need to specify the WebJar version once in the `using dep` directive, this makes manually bumping easier.
- We reduce the boilerplate of loading a file from a WebJar

---

Felt weird doing all this without fs2 or IO, but so far none of our uses of this are in IO.
The whole `org.graalvm.polyglot.Context` stuff feels like it could have a cool little micro library around it.

### DIFF
--- a/build.scala
+++ b/build.scala
@@ -405,13 +405,9 @@ object LaikaCustomizations {
   }
 
   val Icons = {
-    def loadFaIcon(prefix: String, name: String) = {
-      val resourcePath =
-        "/META-INF/resources/webjars/fortawesome__fontawesome-free/7.2.0"
-      val inputStream =
-        getClass.getResourceAsStream(s"$resourcePath/svgs/$prefix/$name.svg")
-      String(inputStream.readAllBytes())
-    }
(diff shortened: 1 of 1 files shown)


build.scala
```diff
@@ -476,6 +470,36 @@ object KaTeX {
 
 }
 
+// Provides access to resources of a bundled WebJar.
+class WebJar(artifactId: String) {
+  private val version: String =
+    val propsPath = s"META-INF/maven/org.webjars.npm/$artifactId/pom.properties"
+    val stream = Thread
+      .currentThread()
+      .getContextClassLoader
```
armanbilge: I feel like the `ClassLoader` should be a property of the `WebJar` class, so that the same one is used consistently each time (and it's not at the whim of the thread it's running on).

valencik: Makes sense. I've replaced these two bits of code with:
```scala
private val classLoader = classOf[WebJar].getClassLoader
```

armanbilge (approved):

PR #637, opened 2026-04-16
Update protosearch, exclude the 404 page

valencik:
Updates to latest protosearch snapshot, which let's us exclude paths and specify our config just once.

**Excluding the 404 page**

Before:
image

After:
image

### DIFF
--- a/build.scala
+++ b/build.scala
@@ -80,8 +80,9 @@ object LaikaBuild {
   import pink.cozydev.protosearch.laika.IndexConfig
   import pink.cozydev.protosearch.ui.SearchUI
 
+  val indexExclusions = Path.Root / "404.md" +: Redirects.paths
   val indexConfig =
-    IndexConfig.withExcludedPaths(Path.Root / "404.md")
+    IndexConfig.withExcludedPaths(indexExclusions*)
 
   def input = {
     val securityPolicy = new URI(
@@ -490,6 +491,8 @@ object Redirects {
   def inputTree = map.foldLeft(InputTree[IO]) { case (tree, (from, to)) =>
     tree.addString(mkRedirect(to), Root / (from.stripSuffix(".html") + ".md"))
   }
+
+  def paths = map.keys.map(p => Root / p.stripSuffix(".html")).toList
 
   private def mkRedirect(to: String) =
     s"""{% laika.html.template = "/templates/redirect.template.html", laika.targetFormats: [html], target = "$to" %}"""
--- a/build.scala
+++ b/build.scala
@@ -4,7 +4,7 @@
 //> using dep org.graalvm.js:js:25.0.2
 //> using dep org.webjars.npm:katex:0.16.44
 //> using dep org.webjars.npm:fortawesome__fontawesome-free:7.2.0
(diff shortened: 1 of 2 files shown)


build.scala
```diff
@@ -77,9 +77,12 @@ object LaikaBuild {
   import laika.io.syntax.*
   import laika.parse.code.languages.ScalaSyntax
   import laika.theme.*
-  import pink.cozydev.protosearch.analysis.{IndexFormat, IndexRendererConfig}
+  import pink.cozydev.protosearch.laika.IndexConfig
   import pink.cozydev.protosearch.ui.SearchUI
 
+  val indexConfig =
+    IndexConfig.withExcludedPaths(Path.Root / "404.md")
```
valencik: I think this part is neat, we can define the config and settings once and then access the BinaryRendererConfig with `indexConfig.config` and the `TwoPhaseRenderFormat` for Transformers with `indexConfig.format`.

src/main.js
```diff
@@ -26,7 +26,7 @@ function hideSearchModal() {
 }
 
 function renderHit(hit) {
-  const link = `${hit.fields.path}.html`
+  const link = `/${hit.fields.path}.html`
```
valencik, line 29: Because the `path` field no longer has the leading slash.
Which means queries like `path:blog` now work

build.scala
```diff
@@ -484,6 +492,8 @@ object Redirects {
     tree.addString(mkRedirect(to), Root / (from.stripSuffix(".html") + ".md"))
   }
 
+  def paths = map.keys.map(p => Root / p.stripSuffix(".html")).toList
```
valencik, line 495: So we can exclude these empty redirect documents from the index.
They all end up with the same name `index.txt` because `index.md` becomes `index.txt` after "plaintext" formatting in the Index renderer.

This is from the live site:

Image

armanbilge (commented):
This is great! But I wonder if it could be better UX to indicate in the config of a page itself  (or a directory) whether it should be indexed or not?

valencik:
> This is great! But I wonder if it could be better UX to indicate in the config of a page itself (or a directory) whether it should be indexed or not?

I agree it should be possible to leverage HOCON config to exclude from the index. That would probably be nice for excluding the 404 page. However, I don't think that would have been any better of a solution for the redirects.

Currently the indexing procedure is ignorant of any HOCON data. I think influencing protosearch indexing via HOCON should be supported in the protosearch laika module itself. There's some WIP-ish stuff here: https://github.com/cozydev-pink/protosearch/pull/347 but that kinda stalled once I realized it wouldn't work for our complex author objects we use here, only for flat/scalar values.


All this to say, I think we should have some HOCON support, but I don't think we should hold up this PR waiting for it.

armanbilge:
> However, I don't think that would have been any better of a solution for the redirects.

It could be, we would just add it to the config here.
https://github.com/typelevel/typelevel.github.com/blob/e579a1c9542615b0a520700fc3fcb955b9ff7d74/build.scala#L509

valencik:
Chatted with Arman about this while pairing.
We're thinking that all of this config could be done in HOCON either at the document level (for the redirects and 404 page for example) or at the directory level (if we wanted to index certain keys in one directory but not another).

PR #638, opened 2026-04-17
Intercept / keystroke only if search modal not active

armanbilge:
Fixes the problem where you cannot type `/` when searching.


valencik (approved):
nice

PR #667, opened 2026-07-15
Fix appearance of sponsors

reardonj:
I realized that I only can see 2 sponsors in Firefox on Mac. A quick googling indicates that this happens when an SVG is missing a width and height. I added attributes matching the view boxes, and now they show up!

While meddling with that, I saw the logos go off the edge of the screen at medium resolutions before the dsiplay is narrow enough that they become one column. I changed the flexbox settings to let it wrap

Was the intention to avoid multiple rows? If so, I guess we'd have to turn it into a column sooner instead.

## Firefox Issue

Before (Firefox):
image

After (Firefox):
image

After (Safari):
image

## Small Display Issue

Before:
image

After:
image


jducoeur (approved):

armanbilge (approved):
oops, thank you!

PR #671, opened 2026-07-23
Show latest blog posts on the home page

reardonj:
This PR adds a section to show the latest blog posts on the typelevel.org homepage, to give it some changing content. This could probably use a little workshopping:

- above or below sponsors? I put it above since it is less static content.
- should there be a link to all blog posts somewhere? I didn't find an option I liked.
- I dropped the cards to keep the layout less noisy on this more complicated page, but that does mean it looks a bit different from the regular blog page. Since there are no boxes, I left-aligned the date to keep the data organized. Similarly, I tweaked the colors

desktop

mobile

### DIFF
--- a/build.scala
+++ b/build.scala
@@ -253,9 +253,8 @@ object LaikaCustomizations {
       TemplateDirectives.eval("forBlogPosts") {
         import TemplateDirectives.dsl.*
 
-          val count = attribute(0).as[Int].optional
-
-          (count, cursor, parsedBody, source).mapN { (count, c, b, s) =>
+        (attribute(0).as[Int].optional, cursor, parsedBody, source)
+          .mapN { (count, c, b, s) =>
             def contentScope(value: ConfigValue) =
               TemplateScope(TemplateSpanSequence(b), value, s)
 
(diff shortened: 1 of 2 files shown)


jducoeur (approved):
I like this a lot in the web layout.  I'm of more mixed minds on the mobile layout, but there's enough content that I'm not sure there is any great answer -- at the least, I think it's a plus.

In terms of a link to all blog articles, you mean one grouped with these new cards?  I'm not sure it's a high priority, given that the main "Blog" link is pretty visible at the top.  There might be an improvement to be had there, but I don't think it's a blocker.

reardonj:
> I like this a lot in the web layout. I'm of more mixed minds on the mobile layout, but there's enough content that I'm not sure there is any great answer -- at the least, I think it's a plus.

Realistically, it's not a site for using on mobile anyways. We're just making a best effort so it isn't unusable on small screens.

> In terms of a link to all blog articles, you mean one grouped with these new cards? I'm not sure it's a high priority, given that the main "Blog" link is pretty visible at the top. There might be an improvement to be had there, but I don't think it's a blocker.

Correct, you often see such a link, but I didn't come up with one I liked. I don't love taking up more vertical space for one.

samspills (approved):

Koroeskohr (approved):

PR #672, opened 2026-07-24
Bump protosearch to latest snapshot

valencik:
Bump protosearch to latest snapshot which leverages a single configuration point so we can reuse the same config for the binary renderers and the preview server.

### DIFF
--- a/build.scala
+++ b/build.scala
@@ -4,7 +4,7 @@
 //> using dep org.graalvm.js:js:25.1.3
 //> using dep org.webjars.npm:katex:0.17.0
 //> using dep org.webjars.npm:fortawesome__fontawesome-free:7.2.0
-//> using dep pink.cozydev::protosearch-laika:0.0-7f79720-SNAPSHOT
+//> using dep pink.cozydev::protosearch-laika:0.0-f6fbb35-SNAPSHOT
 //> using repository https://central.sonatype.com/repository/maven-snapshots
 //> using option -deprecation
 
@@ -77,8 +77,10 @@ object LaikaBuild {
   import laika.io.syntax.*
   import laika.parse.code.languages.ScalaSyntax
   import laika.theme.*
-  import pink.cozydev.protosearch.analysis.{IndexFormat, IndexRendererConfig}
+  import pink.cozydev.protosearch.laika.IndexConfig
   import pink.cozydev.protosearch.ui.SearchUI
+
+  val indexConfig = IndexConfig.default
 
   def input = {
     val securityPolicy = new URI(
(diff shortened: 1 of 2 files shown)


reardonj:
I'm trying this out locally. I have observed that if I search in a subdirectory (eg. `http://localhost:8000/community/`) the search result links are relative to that directory. eg. if I search for "foundation", I get a link to: `http://localhost:8000/community/foundation/index.html`, instead of `http://localhost:8000/foundation/index.html`. The live site behaves correctly.

reardonj:
I was able to fix this locally by modifying https://github.com/typelevel/typelevel.github.com/blob/07bfeca07ccd898e040ce217534d013f6445f86c/src/main.js#L46 to prefix the `href` with a slash

```diff
-        <a class="bulma-subtitle bulma-has-text-link" href="${link}">${title}</a>
+        <a class="bulma-subtitle bulma-has-text-link" href="/${link}">${title}</a>
```

valencik:
Ahh yes, the `path` field now omits the top level slash.
I pulled in the same fix from another PR here that included these changes
https://github.com/typelevel/typelevel.github.com/pull/637/changes#r3090174184

src/main.js
```diff
@@ -32,7 +32,7 @@ function hideSearchModal() {
 }
 
 function renderHit(hit) {
-  const link = `${hit.fields.path}.html`
+  const link = `/${hit.fields.path}.html`
```
valencik, line 35: The indexed `path` field now omits the top level slash.

build.scala
```diff
@@ -77,9 +77,11 @@ object LaikaBuild {
   import laika.io.syntax.*
   import laika.parse.code.languages.ScalaSyntax
   import laika.theme.*
-  import pink.cozydev.protosearch.analysis.{IndexFormat, IndexRendererConfig}
+  import pink.cozydev.protosearch.laika.IndexConfig
   import pink.cozydev.protosearch.ui.SearchUI
 
+  val indexConfig = IndexConfig.default
```
valencik, line 83: The idea here is that we configure the protosearch-laika `IndexConfig` once and can then reuse it to get the Laika `BinaryRendererConfig` or `IndexFormat` depending on which we need.
Currently we're just using the default config, in the future we might configure other settings like which HOCON keys to add to the index as data.

valencik:
I think we should bump up the number of search results from 10 (default) to like 50.
Previously there was no limit in place.
The search API supports pagination now, but there's nothing in place in our UI for it, and that's probably fine for now.

reardonj:
50 sounds like plenty

valencik:
Ready for review again.

The path issue is resolved, being on the community page and searching 'monad' should:
- show only the top 50 results
- links correctly go to blog posts

Additionally folks should be able to build the site locally again even with a clean cache.

Hopefully the last snapshot bump before a 0.0.1 release

jducoeur (approved):

PR #676, opened 2026-07-28
Update Start Contributing Guide and Community pages

reardonj:
@abby-ql , I pulled out a few more updates I was making to the start contributing page to it's own PR while I finish off the rest of my work. 

Later this week I'll have another PR with more detailed contributing guidelines covering details like copyright and other stuff the GSoC guidelines had, but they're quite separate from these updates.

### reorganized the community page

- make layout take up move space on wide screens. Same behavior on smaller screens. With the pages already having 'columns' letting it take up a bit more space saves users on large monitors having to scroll as much to see everything.
- expand descriptions, so the new sections are similar sized to the others
- move new sections ahead of coc. I like resources and code of conduct last, since they're not quite get involved pages like the others

> [!NOTE]
> I feel there should be a way to template the community page instead of having to update all these css classes, but I didn't want to go down a bikeshedding rabbit-hole.

community page

### tweaks to the start contributing guide

- don't say it's simple, it's a phrase that tends to make newcomers assume the opposite. 
- I added links to all the suggestions and put them as a 'Become a user of the ecosystem you want to contribute to.' section
- use bold instead of incorrect subheading levels. 
- add AI note

image

### DIFF
--- a/src/community/README.md
+++ b/src/community/README.md
@@ -26,9 +26,9 @@ The Typelevel community is dedicated to providing a positive experience for ever
 @:@
 
 @:fragment(contributing)
-Learn how to get involved in Typelevel projects. Find good first issues, understand the workflow, and collaborate with maintainers.
+The projects in the Typelevel ecosystem are all open-source. Learn how to contribute to Typelevel and help maintain your favorite projects. Find good first issues, understand the workflow, and collaborate with maintainers.
 @:@
 
 @:fragment(resources)
-Explore recommended books, courses, and documentation for learning functional programming in Scala and the Typelevel ecosystem.
+Learn functional programming fundamentals and the theory behind the design of Typelevel libraries. Explore recommended books, courses, and documentation for learning functional programming in Scala and the Typelevel ecosystem.
 @:@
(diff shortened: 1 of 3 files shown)


armanbilge:
When I set up the site I also had a vision of adding a "Get Started" button and "Learn" menu item to the navbar, but lacking content at the time. So we could consider bringing those to life.

https://github.com/typelevel/typelevel.github.com/blob/da3dca5a77ff54d5b75943727609d067787f4caa/src/templates/nav.template.html#L20-L21

reardonj:
> When I set up the site I also had a vision of adding a "Get Started" button and "Learn" menu item to the navbar, but lacking content at the time. So we could consider bringing those to life.
> 
> https://github.com/typelevel/typelevel.github.com/blob/da3dca5a77ff54d5b75943727609d067787f4caa/src/templates/nav.template.html#L20-L21

oooo. I think `Learn` would be better pointed at the future TL learning guide, but get started could point to @abby-ql 's page

reardonj:
But I think it best to turn that on once we have the full contributing guide too.

jducoeur (approved):

PR #677, opened 2026-07-30
Initial prototype of interactive tutorials

jducoeur:
Inspired by ScalaTutorials, this is a largely vibe-coded prototype of how we might do something similar on the TL website.

This is very intentionally *not* surfaced to the rest of the site yet: since it is just a prototype, I'd like us to be able to iterate on it before we declare it good enough to open up to the public.

A small constraint: due to a quirk of how Scastie works, if you are working in this locally, do *not* try to navigate to it using `localhost`. Go explicitly to http://127.0.0.1:8000/tutorials/

This prototype demonstrates the idea of having multiple "tracks" of tutorial -- I'm pretty sure that, if we go down this road, we'll want to do that. But the actual content here is entirely machine-written, and should be considered entirely throwaway: I just wanted enough content to be able to show off what I was thinking.

None of this should be considered a statement that we are necessarily going to do this.  But I figured it was worth throwing something together to show that it's possible, and we can then discuss whether we want it, and begin to iterate on the details if so.

As it says, this first draft is almost 100% vibe-coded.  See the first commit for the full prompt.

Screenshot 2026-07-30 at 6 34 07 PM

Screenshot 2026-07-30 at 6 34 28 PM

Screenshot 2026-07-30 at 6 35 02 PM

### DIFF
--- a/src/tutorials/README.md
+++ b/src/tutorials/README.md
@@ -4,7 +4,7 @@
 
 # Typelevel Tutorials
 
-Welcome! These tutorials are hands-on: each lesson explains an idea on this side
+Welcome! These tutorials are hands-on: each lesson explains an idea on the left side
 of the page and gives you a live Scala editor on the other. Edit the code, press
 **Run**, and see what happens.
 
--- a/src/tutorials/README.md
+++ b/src/tutorials/README.md
@@ -1,0 +1,37 @@
+{%
+  laika.title = Tutorials
+%}
+
+# Typelevel Tutorials
+
+Welcome! These tutorials are hands-on: each lesson explains an idea on this side
+of the page and gives you a live Scala editor on the other. Edit the code, press
+**Run**, and see what happens.
+
+@:style(bulma-notification bulma-is-warning bulma-is-light)
+  This section is an early prototype. The lessons below exist to show that the
+  machinery works; their content will be replaced.
+@:@
+
+## Cats
+
+[Cats](https://typelevel.org/cats/) is the foundation of the Typelevel
+ecosystem: a library of type classes and data types for functional programming
+in Scala.
+
+@:navigationTree {
+  entries = [
+    { target = "/tutorials/cats", excludeRoot = true, excludeSections = true }
+  ]
+}
+
+## Cats Effect
+
+[Cats Effect](https://typelevel.org/cats-effect/) builds on Cats to describe
+side effects as values, and gives you a runtime to execute them.
+
+@:navigationTree {
+  entries = [
+    { target = "/tutorials/cats-effect", excludeRoot = true, excludeSections = true }
+  ]
+}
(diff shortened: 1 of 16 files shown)


reardonj:
Cool. For the TL site, maybe getting them inline on the page would be even better for some brief runnable samples. The whole tutorial format might be better for the actual project sites.

dhinojosa:
How did AI know so much about Laika all of a sudden? I tried to get it to help me before and it kept giving out wrong information.

reardonj:
> How did AI know so much about Laika all of a sudden? I tried to get it to help me before and it kept giving out wrong information.

In my experience, if you point them at a repository full of sample code, it manages to follow the patterns reasonably well.

jducoeur:
Yeah -- Opus, especially, is increasingly good at just following my lead.  That's the only reason I'm using it at work: it's doing nicely at following my particular TL-centric style (with my own idiosyncracies) and building code that looks broadly similar.

jducoeur:
> For the TL site, maybe getting them inline on the page would be even better for some brief runnable samples.

Possibly, but one problem at a time IMO.  I'm not sure how well the embedded-Scastie affordances will work from a UX POV in such a small space (has anyone pioneered that yet?), and I'd prefer we not make the best the enemy of the good.

> The whole tutorial format might be better for the actual project sites.

Agreed, but I'm not sure the difference matters all *that* much in practical terms, now that the projects are increasingly served from the typelevel.org domain.  That is, if Cats is located at typelevel.org/cats, and its tutorial is at typelevel.org/tutorials/cats, that's not a surprise from the user POV.  And I deliberately broke it down to one-directory-per-track, specifically so that it's easy for us to keep it all straight in the data.

It's arguably a little easier *organizationally* for these to belong to the individual project repos.  But *techically*, I suspect it would be more of a pain in the ass, especially since the projects aren't necessarily unified in how they're serving their documentation: replicating this code in each project's site feels like a lose.

My suspicion is that it would make sense for this to be served individually from project sites if and only if we built this functionality *into* Laika.  That's a somewhat plausible idea (arguably even a great one once Scala can compile in-browser), but again, I'd probably argue that it's best as a later step unless someone is really motivated to spearhead it.

All that said, if we do go this route, I'd recommend we throw the doors open and say that all Organization projects (and *maybe* Affiliate projects, although that's not as obvious) are allowed and encouraged to contribute tracks for their projects, and all we need to do is accept PRs.  It's the sort of thing that I'd say the TSC should facilitate, but not necessarily do all the content.

reardonj:
Ahh, alright, this plan makes sense.

src/tutorials/README.md
```diff
@@ -35,3 +35,7 @@ side effects as values, and gives you a runtime to execute them.
     { target = "/tutorials/cats-effect", excludeRoot = true, excludeSections = true }
   ]
 }
+
+These tutorials were inspired by [scalatutorials](https://scalatutorials.com), which teaches
```
reardonj, line 40: If you really want to recommend them, there is also that new learning resources page, which only has one link under tutorials and courses right now!

jducoeur: Oh, good point -- would be totally appropriate to have there IMO.  (Although probably in a separate PR, since it's not really related to this prototype.)

PR #684, opened 2026-08-13
Update release job to use `upload-pages-artifact`

### LINKED ISSUE #658: Update/upgrade deploy workflow
I think the `peaceiris` action has fallen out of maintenance. We should investigate the official [`upload-pages-artifact`](https://github.com/actions/upload-pages-artifact) action.

https://github.com/typelevel/typelevel.github.com/blob/3907b3643d5461ecca7b4d4a31e16ac7f51a6c71/.github/workflows/ci.yml#L25-L30

Koroeskohr:
I implemented all of this blind without knowing how to properly test it.

I think this now makes the usage of the gh-pages branch obsolete, we can think about deleting it if this all works out. 

Used AI only to figure out what permissions I would need.

Fixes #658

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,9 @@ on:
   push:
     branches: ["main"]
 
+permissions:
+  contents: read
+
 jobs:
   build:
     name: Build and Test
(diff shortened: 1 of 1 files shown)


reardonj (commented):
This looks similar (but slightly more up to date) than my blog's setup, so hopefully it's right.

PR #689, opened 2026-09-07
Add Typelevel welcome guide

dhinojosa:
Add an introductory welcome guide to Typelevel

 The guide begins with familiar Scala concepts and gradually introduces:

  - Typelevel's mission and community
  - Representational types and making illegal states unrepresentable
  - Immutability and function composition
  - `map`, `filter`, `flatMap`, and for-comprehensions
  - Side effects and referential transparency
  - The motivation behind effect types
  - A small `IO`-inspired data type built step by step
  - Programs, algebras, and interpreters
  - An overview of the Typelevel ecosystem

  The page is intended for readers without prior functional programming experience. It includes runnable Scastie examples and diagrams to make the concepts more
  approachable.


SethTisue:
This will be an excellent addition to the site! Super useful.

But I do have a suggestion on structure:

After the introductory paragraphs about community, the transition into specific technical explanations (of types and `flatMap` and such) is rather abrupt and disorienting.

Before you launch into talking about what types are, I'd suggest inserting a section that summarizes what's coming. The gist of the added material would be: "In the main body of this  document, we'll give you a crash course on what typed functional programming is. We'll explain why types and immutability are so important and beneficial to software quality, and we'll show how they're expressed in Scala. We'll introduce you to the `IO` type, [and so on]... Finally, at the end of the document, we'll introduce you to the libraries that make up the Typelevel ecosystem..." that kind of thing. So the reader will know what's coming so they can continue reading or not, skip ahead or not. Even if they decide to plow straight through and read the whole thing, knowing the bigger picture will help them stay oriented as they read.

(I'm not sure how lengthy the transition should be. Perhaps you'll decide some brief transitional material is enough, or perhaps you'll want to dwell it on longer...?)

dhinojosa:
Noice! Appreciate it @SethTisue 

The 3rd to 5th paragraphs have that
"This introduction starts with ordinary Scala values and functions. From there, we will learn how map, flatMap, and for-comprehensions let us combine values without discarding the information carried by their types.

Finally, we will use those ideas to build a small version of IO, the foundational data type from Cats Effect.

You do not need prior experience with functional programming. Every new idea builds on the previous one."

But I see your point that there is a (or more) spots where it feels like it goes from hot water to cold water. Let me see what I can do to ease that transition

dhinojosa:
https://pr-689.typelevel-website.pages.dev/welcome/welcome is where the page resides, by the way. I have not linked it to the front page.

src/welcome/welcome.md
```diff
@@ -0,0 +1,1354 @@
+# Introduction to Typelevel
+
+## What is Typelevel?
+
+Typelevel is a non-profit organization dedicated to providing principled, type-safe, functional programming tools for the Scala ecosystem. We are hyper-focused on functional programming and building a welcoming, inclusive community around pure functional programming in Scala.
```
jducoeur, line 5: I might tone that down to just "focused" - "hyper-focused" feels like over-egging the pudding a little bit.  (Especially in that we *also* try to be very pragmatic about what actually works, and things aren't necessarily quite *that* pure under the hood.)

src/welcome/welcome.md
```diff
@@ -0,0 +1,1354 @@
+# Introduction to Typelevel
+
+## What is Typelevel?
```
jducoeur, line 3: I feel like this section is combining two rather different purposes: introducing what Typelevel-the-organization is, and starting to introduce the practical tutorial about coding within our ecosystem.  The overall flow winds up feeling a little disjoint, especially with the "Welcoming Community" bit coming in between a few paragraphs about the technical stuff and then diving into it more deeply.

I'd probably recommend rearranging slightly -- shortening this intro section to just mention the tutorial that makes up the bulk of the page, and moving the last three paragraphs down into a section of their own, as the beginning of a top-level "Tutorial" section.

src/welcome/welcome.md
```diff
@@ -0,0 +1,1354 @@
...
+collaboration, knowledge sharing, and mentorship through GitHub, Discord, community events, and conferences.
+
+Our [Code of Conduct](https://typelevel.org/code-of-conduct/) describes the standards we expect in every Typelevel space:
+be kind and patient, assume good intentions, offer feedback constructively, and respect different experiences and
+viewpoints. Whether you are writing your first Scala program or maintaining a functional programming library, you belong
+here.
+
+## Let Types Show Your Work
+
+Types are an essential part of Scala and central to the philosophy of Typelevel. If you think back to your professor or teacher in mathematics and physics, they would tell you to “label your work”. A value of `10` could represent 10 meters, 10 seconds, 10 kilograms, or 10 meters per second. The labels provide context and help verify that your calculations are correct. If you expected a result in meters per second but instead got kilometers per second, the labels immediately indicated that something had gone wrong.
```
jducoeur, line 29: > If you think back to your professor or teacher in mathematics and physics

Musing: should we worry about this being classist?  I think it's correct for the typical programmer, but I worry about it coming across as snooty to the self-taught folks who didn't come up through STEM.

src/welcome/welcome.md
```diff
@@ -0,0 +1,1354 @@
...
+
+Programming follows the same principle. Rather than relying on comments or variable names to describe what a value represents, we use types. A type gives meaning to a value, defines which values are valid, and determines which operations make sense. This philosophy is at the heart of the Typelevel ecosystem.
```
jducoeur, line 31: > defines which values are valid

Might be worth leaning into this a little bit more, since it's the heart and soul of strongly-typed programming IMO, and wildly unobvious to folks coming from dynamically-typed languages.  The notion that a type is synonymous with the set of values that you can put into a variable or parameter is an important intuition to impart.

Even more, the critical point is that a type defines what values *can't* be placed here, and why that's so enormously helpful.  Folks coming from a non-compiled languages tend to complain about the compilation process, and strong types are the baseline of why the compiler is really helpful.  So in your power example below, it might be helpful to show an error example of trying to combine types illegally, and how the compiler prevents that mistake.

src/welcome/welcome.md
```diff
@@ -0,0 +1,1354 @@
...
+loadKw: Double = 1.2
+
+scala> val energyMwh = 24.2
+energyMwh: Double = 24.2
+
+scala> val sumKw = loadKw + energyMwh
+sumKw: Double = 25.4
+```
+
+In the example above, what stops us from doing this when we use plain numbers? One is a Kilowatt, the other is Megawatt-Hours, and we were attempting to add two numbers that really shouldn’t be added together. Using raw numbers like this is a code smell called _primitive obsession_. We are obsessing over `Double` in this case for everything! This is why we have to show our work with labels in our math and physics homework; the same goes for our programming language. A better option is to use a wrapper around the number; this is called a _value object_.
```
jducoeur, line 47: > this is called a _value object_.

Is it?  I usually think of this as "value class" instead, although I'm not certain what's more common.

src/welcome/welcome.md
```diff
@@ -0,0 +1,1354 @@
...
+scala> val sum = load1 + load2
+sum: squants.energy.Power = 35.0 kW
+
+scala> sum == Kilowatts(35)
+res0: Boolean = true
+
+scala> sum == Megawatts(0.035) // comparisons automatically convert scale
+res1: Boolean = true
+```
+**Try it in Scastie!** [![Try it in Scastie](../img/media/welcome/scastie.png)](https://scastie.scala-lang.org/5SNhXYrTRxWFp8lR10da7A)
```
jducoeur, line 65: Not a blocker, but down the line we might want to investigate whether we can inline the Scastie window usefully -- that might reduce friction a little.

(If they manage to get in-browser Scastie working, I'd say that becomes a slam-dunk; as it is, it's worth thinking about.)

jducoeur: Separately: clicking on this appears to replace the current tab with Scastie, which is a roadbump.  I'm not certain whether this Markdown dialect has an easy way to do that as a new tab, but that would be strongly preferable IMO.

src/welcome/welcome.md
```diff
@@ -0,0 +1,1354 @@
...
+list is never empty,” we can use `NonEmptyList` from [Cats](https://github.com/typelevel/cats):
+
+```scala
+import cats.data.NonEmptyList
+
+def average(xs: NonEmptyList[Int]): Double =
+  xs.reduceLeft(_ + _) / xs.length.toDouble
+```
+
+The type tells callers that `average` always has at least one value to work with. Its methods preserve that guarantee when
```
jducoeur, line 81:
```suggestion
The type tells callers that `average` requires at least one value to work with. Its methods preserve that guarantee when
```

src/welcome/welcome.md
```diff
@@ -0,0 +1,1354 @@
...
+Whether we are modeling domain types such as `Human` and `Kilowatts` or working with collections such as `List` and `NonEmptyList`, we generally prefer immutable values. Immutability makes values and their state easier to reason about. Immutable values can also be shared safely between threads without coordinating access to changing state, reducing the need for locks, synchronization, and defensive copying.
+
+Immutability is also a core design principle in functional programming and is strongly emphasized throughout the Typelevel ecosystem.
+
+A `case class` is one typical way that we make an immutable class. 
+
+```scala
+final case class User(id: Long, name: String, favoriteLanguageId: Long)
+final case class Language(id: Long, name: String)
+```
```
jducoeur, line 125: Very minor note, but I'm concerned that a Scala novice might look at this section and conclude that `final` has something to do with immutability.  I'd probably drop that in this example: while it's true that case classes shouldn't generally be extended (so `final` is a nice little belt-and-suspenders protection), it's minor (I essentially never bother in my production systems), and I think it's a distraction from your main point here.

jducoeur (commented):
Took a little while to begin a deep read of the text.  I'm only about a quarter of the way through, but it's time to go do other things, so here are a few thoughts so far.  Nothing critical, but some possibilities to consider.

src/welcome/welcome.md
dhinojosa: Sounds good.

src/welcome/welcome.md
dhinojosa: I think I got it per @SethTisue's recommendations

src/welcome/welcome.md
dhinojosa: Great I will update here in the next push

src/welcome/welcome.md
dhinojosa: Unsure on this one; it wouldn't be too much of an assumption that the reader took some sort of math and science even in secondary school. Right?

src/welcome/welcome.md
dhinojosa: That's a good question. I have always used "value object" since that is the Domain-Driven Design definition and how I have used it for years.

Martin Fowler: "If I have two point objects that both represent the Cartesian coordinates of (2,3), it makes sense to treat them as equal. Objects that are equal due to the value of their properties, in this case their x and y coordinates, are called value objects."

https://martinfowler.com/bliki/ValueObject.html

[JEP 401 is proposed as a preview feature for JDK](https://openjdk.org/jeps/401); there will be a new feature that would strip away the object wrapper that will "distinguish objects by the values of their fields rather than their identities". They use "value object" here.

How I am interpreting that is what is already happening in the Scala space when we do use `extends AnyVal`, called value types or value classes, or we use opaque types for Scala 3.

https://docs.scala-lang.org/overviews/core/value-classes.html
https://docs.scala-lang.org/scala3/book/types-opaque-types.html

This Venn Diagram does some overlapping for sure. 

I suppose then I am using value object as the design term, whereas a value class is a Scala technical mechanism for potentially avoiding the allocation of a wrapper object. An `opaque` type is the recommended Scala 3 mechanism for getting a distinct compile-time type without adding a runtime wrapper.

jducoeur: Hmm.  I get the intent, but I *think* this terminology is a Java-ism?  I don't recall coming across that specific phrasing in the Scala community, which is why this is striking my ear weirdly.  It's not necessarily wrong, but "this is called" feels overstated to me in this context.

(And opaque types specifically *aren't* what you're calling a value object -- they're straight-up aliases, albeit aliases with firmer walls than usual.)

src/welcome/welcome.md
dhinojosa: I would have to lean on a Laika directive or find out a way to do `target-"blank"`

src/welcome/welcome.md
dhinojosa: Done

src/welcome/welcome.md
dhinojosa: Agree, changed.

dhinojosa:
> Before you launch into talking about what types are, I'd suggest inserting a section that summarizes what's coming. The gist of the added material would be: "In the main body of this document, we'll give you a crash course on what typed functional programming is. We'll explain why types and immutability are so important and beneficial to software quality, and we'll show how they're expressed in Scala. We'll introduce you to the `IO` type, [and so on]... Finally, at the end of the document, we'll introduce you to the libraries that make up the Typelevel ecosystem..." that kind of thing. So the reader will know what's coming so they can continue reading or not, skip ahead or not. Even if they decide to plow straight through and read the whole thing, knowing the bigger picture will help them stay oriented as they read.
> 
> (I'm not sure how lengthy the transition should be. Perhaps you'll decide some brief transitional material is enough, or perhaps you'll want to dwell it on longer...?)

I changed the intro to the following and will commit momentarily. @SethTisue 

# Introduction to Typelevel

## What is Typelevel?

Typelevel is a non-profit organization dedicated to providing principled, type-safe, functional programming tools for the Scala ecosystem. We are hyper-focused on functional programming and building a welcoming, inclusive community around pure functional programming in Scala.

Our mission is to enable the creation of robust, maintainable software through functional programming. We develop and maintain a suite of libraries that help you write safer, more composable, and more testable code.

You do not need prior experience with functional programming. Every new idea builds on the previous one.

## What to Expect

The main body of this introduction is a orientation in typed functional programming as Typelevel practices it in Scala. We will start with types: how they give meaning to values, prevent invalid states, and let the compiler help us check our work. From there, we will look at immutability, small functions, and the basic tools Scala gives us for transforming and composing values, including `map`, `flatMap`, and for-comprehensions.

Those ideas will lead us into side effects, referential transparency, and the question every useful program eventually faces: how do we read from the world, write to the world, or generate random values without giving up the benefits of pure functional programming? To answer that, we will build a small teaching version of `IO`, showing how effectful programs can be described as data and interpreted later.

Near the end, we will connect those ideas to common Typelevel terms such as programs, algebras, and interpreters. Finally, we will introduce the libraries that make up the Typelevel ecosystem, including Cats, Cats Effect, FS2, http4s, Circe, Doobie, and Skunk, so you can see where each piece fits and decide what to explore next.

## A Welcoming Community

Building great software goes hand in hand with building a great community. Typelevel is committed to an inclusive,
welcoming, and safe setting for everyone, regardless of background, identity, or experience level. We encourage
collaboration, knowledge sharing, and mentorship through GitHub, Discord, community events, and conferences.

SethTisue:
LGTM

src/welcome/welcome.md
dhinojosa: Simmering on this...

src/welcome/welcome.md
dhinojosa: Booyeah, it works now. I just used the `@:@html` directive.

https://pr-689.typelevel-website.pages.dev/welcome/welcome

I fixed some images and the LateX rendering along the way

